PR #4, opened 2021-03-20
Allow use of Zygote or ReverseDiff

mohamed82008:
This PR enables the use of Zygote or ReverseDiff in ADNLPModel using Requires.jl to avoid depending on either. Reverse-mode AD would unlock the true power of the augmented Lagrangian algorithm in Percival.jl since it avoids the computation of the whole Jacobian.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,4 +23,4 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 
 [targets]
-test = ["NLPModelsModifiers", "NLPModelsTest", "Test", "Zygote", "ReverseDiff"]
+test = ["NLPModelsModifiers", "NLPModelsTest", "ReverseDiff", "Test", "Zygote"]
--- a/Project.toml
+++ b/Project.toml
@@ -6,17 +6,21 @@ version = "0.1.0"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NLPModels = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 ForwardDiff = "0.9.0, 0.10.0"
 NLPModels = "0.14"
+Requires = "1"
 julia = "^1.3.0"
 
 [extras]
 NLPModelsModifiers = "e01155f1-5c6f-4375-a9d8-616dd036575f"
 NLPModelsTest = "7998695d-6960-4d3a-85c4-e1bceb8cd856"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 
 [targets]
-test = ["NLPModelsModifiers", "NLPModelsTest", "Test"]
+test = ["NLPModelsModifiers", "NLPModelsTest", "Test", "Zygote", "ReverseDiff"]
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # ADNLPModels
 
-This package provides a very simple model implement the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API.
+This package provides a very simple model implement the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API. The following packages are supported:
+- `ForwardDiff.jl`: default choice.
+- `Zygote.jl`: you must load `Zygote.jl` separately and pass `ADNLPModels.ZygoteAD()` as the `adbackend` keyword argument to the `ADNLPModel` or `ADNLSModel` constructor.
+- `ReverseDiff.jl`: you must load `ReverseDiff.jl` separately and pass `ADNLPModels.ReverseDiffAD()` as the `adbackend` keyword argument to the `ADNLPModel` or `ADNLSModel` constructor.
 
 ## How to Cite
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ADNLPModels
 
-This package provides a very simple model implement the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API. The following packages are supported:
+This package provides AD-based model implementations that conform to the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API. The following packages are supported:
 - `ForwardDiff.jl`: default choice.
 - `Zygote.jl`: you must load `Zygote.jl` separately and pass `ADNLPModels.ZygoteAD()` as the `adbackend` keyword argument to the `ADNLPModel` or `ADNLSModel` constructor.
 - `ReverseDiff.jl`: you must load `ReverseDiff.jl` separately and pass `ADNLPModels.ReverseDiffAD()` as the `adbackend` keyword argument to the `ADNLPModel` or `ADNLSModel` constructor.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # ADNLPModels
 
 This package provides a very simple model implement the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API.
-It uses [`ForwardDiff`](https://github.com/JuliaDiff/ForwardDiff.jl) to compute the derivatives, which produces dense matrices, so it isn't very efficient for larger problems.
 
 ## How to Cite
 
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -1,0 +1,74 @@
+abstract type ADBackend end
+struct ForwardDiffAD <: ADBackend end
+struct ZygoteAD <: ADBackend end
+struct ReverseDiffAD <: ADBackend end
+
+throw_error(b) = throw(ArgumentError("The AD backend $b is not loaded. Please load the corresponding AD package."))
+gradient(b::ADBackend, ::Any, ::Any) = throw_error(b)
+gradient!(b::ADBackend, ::Any, ::Any, ::Any) = throw_error(b)
+jacobian(b::ADBackend, ::Any, ::Any) = throw_error(b)
+hessian(b::ADBackend, ::Any, ::Any) = throw_error(b)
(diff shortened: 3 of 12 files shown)


Project.toml
```diff
@@ -6,17 +6,21 @@ version = "0.1.0"
...
 [extras]
 NLPModelsModifiers = "e01155f1-5c6f-4375-a9d8-616dd036575f"
 NLPModelsTest = "7998695d-6960-4d3a-85c4-e1bceb8cd856"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 
 [targets]
-test = ["NLPModelsModifiers", "NLPModelsTest", "Test"]
+test = ["NLPModelsModifiers", "NLPModelsTest", "Test", "Zygote", "ReverseDiff"]
```
abelsiqueira, line 26: Can you add these sorted?

dpo:
Thanks! I think that’s a great idea. Do you think it would be possible to also enable ModelingToolkit like this (not necessarily in this PR)?

mohamed82008:
>Do you think it would be possible to also enable ModelingToolkit like this (not necessarily in this PR)?

What do you mean by enable MTK?

test/nls/basic.jl
```diff
@@ -1,34 +1,37 @@
 function autodiff_nls_test()
-  @testset "autodiff_nls_test" begin
-    F(x) = [x[1] - 1; x[2] - x[1]^2]
-    nls = ADNLSModel(F, zeros(2), 2)
+  for adbackend in [
+    ForwardDiffAD(), ZygoteAD(), ReverseDiffAD(),
+  ]
+    @testset "autodiff_nls_test" begin
```
abelsiqueira, line 5:
```suggestion
    @testset "autodiff_nls_test for $adbackend" begin
```

mohamed82008:
Ah for symbolic AD. Yes it would be possible.

dpo:
I mean to compute derivatives.

README.md
```diff
@@ -1,7 +1,6 @@
 # ADNLPModels
 
 This package provides a very simple model implement the [NLPModels](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl) API.
-It uses [`ForwardDiff`](https://github.com/JuliaDiff/ForwardDiff.jl) to compute the derivatives, which produces dense matrices, so it isn't very efficient for larger problems.
```
abelsiqueira, line 4: We should probably add a few lines on the available options for AD.

mohamed82008:
We were a second apart!

abelsiqueira (changes requested):
Thanks, this is a great improvement. I haven't looked into the details of `ad.jl` but overall it looks good. I have a few more comments below:

You also need to add these tests to `nl{p,s}/nlpmodelstest.jl`, which is where the consistency checks are made.

I liked the idea of using struct for the AD backends - kind of like traits. I am wondering though if it makes sense to make `ADNLPModel` parametric and have the type be the AD backend. This could help if some API function can be specialized, though I'm not sure it that's necessary.

Finally, I'll let @amontoison and @tmigot take a look, because they are implementing a new sparse AD model too.

Project.toml
mohamed82008: Sure

mohamed82008:
I removed the Zygote nls models tests because some of these models had mutation and Zygote doesn't support mutation. I think the only failing inverse dependency tests are those of QuadraticModels.jl because of Requires. https://github.com/JuliaSmoothOptimizers/QuadraticModels.jl/pull/35 should fix it.

mohamed82008:
The tests seem to pass now. This is ready for another review.

src/nls.jl
```diff
@@ -148,23 +149,23 @@ function NLPModels.jprod_residual!(nls :: ADNLSModel, x :: AbstractVector, v ::
...
+  Jtv .= pullback(nls.adbackend, nls.F, x, v)
   return Jtv
 end
 
 function NLPModels.hess_residual(nls :: ADNLSModel, x :: AbstractVector, v :: AbstractVector)
   @lencheck nls.meta.nvar x
   @lencheck nls.nls_meta.nequ v
   increment!(nls, :neval_hess_residual)
-  return tril(ForwardDiff.jacobian(x -> ForwardDiff.jacobian(nls.F, x)' * v, x))
+  return tril(jacobian(nls.adbackend, x -> pullback(nls.adbackend, nls.F, x, v), x))
```
abelsiqueira, line 168: You mentioned that `ZygoteAD` fails on the consistency checks, so you removed it. It looks like it's because of this line and the other `hess_coord_residual` below. I changed to the following:
```suggestion
  ϕ(x) = dot(nls.F(x), v)
  return tril(hessian(nls.adbackend, ϕ, x))
```
as similarly to `hess_coord_residual`.
I don't know enough about ADs to know if this breaks something else.

abelsiqueira (changes requested):
Took some more time because I was testing the modifications I proposed.
By the way, when I merge this I intend to squash the commits to keep the history concise. Do you prefer to rearrange and squash them in some other way?

mohamed82008:
>By the way, when I merge this I intend to squash the commits to keep the history concise. Do you prefer to rearrange and squash them in some other way?

Ya feel free to squash.

src/nls.jl
mohamed82008: ya this was it, thanks!

mohamed82008:
Tests seem to be passing.

test/runtests.jl
```diff
@@ -1,4 +1,13 @@
 using ADNLPModels, LinearAlgebra, NLPModels, NLPModelsModifiers, NLPModelsTest, Test
+using Zygote, ReverseDiff
+using ADNLPModels: ForwardDiffAD, ZygoteAD, ReverseDiffAD
+
+function switch_adbackend(nlp::ADNLPModel, adbackend)
+  ADNLPModel(nlp.meta, nlp.counters, adbackend, nlp.f, nlp.c)
```
tmigot, line 6: Since `ADNLPModel` is a mutable struct, couldn't we do `nlp.adbackend = adbackend`? (same for nls)

src/ad.jl
```diff
@@ -0,0 +1,75 @@
+abstract type ADBackend end
+struct ForwardDiffAD <: ADBackend end
+struct ZygoteAD <: ADBackend end
+struct ReverseDiffAD <: ADBackend end
+
+throw_error(b) = throw(ArgumentError("The AD backend $b is not loaded. Please load the corresponding AD package."))
+gradient(b::ADBackend, ::Any, ::Any) = throw_error(b)
```
tmigot, line 7: Maybe, we are now probably missing `@test_throws` type tests for these errors.

tmigot (approved):
Added some comments, but overall I think it is great!

Using `Require.jl` is a neat way of not adding tons of dependencies. However, the drawback is that we have no compatibility control with these packages. Any suggestions for that?

tmigot:
I am personally not a huge fan of `pushforward` and `pullback` as it is not clear they are related to the jacobian. Besides, functions like `pullback` are also used in Zygote to compute the gradient. For instance, what about using `∇c` and `∇ct`?

@amontoison We could very simply fit in this new framework by adding attributes in the `ADBackend` structures with specialized constructors.

mohamed82008:
>I am personally not a huge fan of pushforward and pullback as it is not clear they are related to the jacobian. Besides, functions like pullback are also used in Zygote to compute the gradient. For instance, what about using ∇c and ∇ct?

These are terms from the autodiff world (https://juliadiff.org/ChainRulesCore.jl/stable/#The-propagators:-pushforward-and-pullback). Zygote is the one confusing everyone actually because it uses the term in a different way to return the function output and the pullback function rather than performing the pullback operation. I can rename them to `Jprod` and `Jtprod` if you prefer those.

>Using Require.jl is a neat way of not adding tons of dependencies. However, the drawback is that we have no compatibility control with these packages. Any suggestions for that?

You can have more packages in the compat section than the main dependencies I think. But the API of these packages has been stable for a while so I suspect we need to do much in that regard.

dpo:
Let's use Jprod and Jtprod since that's closer to the optimization nomenclature. Let's also make sure anything that could be `Require`d is tested in the unit tests.

mohamed82008:
Sorry for the late response. I addressed all the comments.

Project.toml
```diff
@@ -6,17 +6,21 @@ version = "0.1.0"
...
 NLPModels = "0.14"
+Requires = "1"
 julia = "^1.3.0"
 
 [extras]
 NLPModelsModifiers = "e01155f1-5c6f-4375-a9d8-616dd036575f"
 NLPModelsTest = "7998695d-6960-4d3a-85c4-e1bceb8cd856"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
```
dpo, line 23: Will this always test the latest release? Is there a way to set versions here?

dpo (changes requested):
Looks amazing, many thanks!!! Just a couple of comments.

Project.toml
mohamed82008: It will use the latest release. I don't recommend setting bounds on optional dependencies. It's possible though.

abelsiqueira:
There's some violent error on Cirrus CI on 1.6:
```
ERROR: Package ADNLPModels errored during testing (received signal: KILL)
```
A similar one in 1.3, with the signal information. They happen in different places.
@dpo, any suggestions? Remove Cirrus CI?

Project.toml
abelsiqueira: > I don't recommend setting bounds on optional dependencies.

Can you expand on this? I am hoping CompatHelper will update them.

dpo:
They happen in the same place: Julia hangs in the constructors for NLS tests. The question is why.

abelsiqueira:
I don't think so, they start after `nls/basic.jl` is run, and inside the NLP consistency.

On [1.3](https://cirrus-ci.com/task/5932749627326464?logs=test#L36), it errors inside the first backend, but on [1.6](https://cirrus-ci.com/task/5574802019713024?logs=test#L94), it errors inside the second backend. Unfortunately, there is no information on where.

Project.toml
mohamed82008: It's just unnecessary bounds. These bounds can create conflicts just because a package exists even if the user doesn't need to load it with ADNLPModels.

abelsiqueira: Makes sense.

dpo:
Could you please rebase, so we can verify that the unit tests pass on FreeBSD?

dpo:
Hmm. I wonder what the difference is between this PR and #7. Here, Julia nightly gets killed on FreeBSD (signal 11).

dpo (approved):

abelsiqueira:
Thanks

PR #8, opened 2021-04-12
add exclude in tests

tmigot:
- remove `exclude_hess` following [NLPModelsTest.jl PR #5](https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/pull/5)
- set `exclude=[]` as ADNLPModels implements all the functions.
- MGH01 also test `FeasibilityResidual` which didn't implement `hess`, `hess_coord` and `ghjvprod`.
- LLS doesn't have `hess` and `hess_coord`.


abelsiqueira:
Thanks

PR #14, opened 2021-04-30
Externalize dense hessian

tmigot:
This PR prepares the field for sparse construction of the Jacobian/Hessian matrix.
- `hess_coord` and `hess_structure` are backend specific
- `nnzj` and `nnzh` are attributes of the backend structure (later on we could also add precompiled Tape ...)

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -1,7 +1,49 @@
 abstract type ADBackend end
-struct ForwardDiffAD <: ADBackend end
-struct ZygoteAD <: ADBackend end
-struct ReverseDiffAD <: ADBackend end
+struct ForwardDiffAD <: ADBackend
+  nnzh::Int
+  nnzj::Int
+end
+function ForwardDiffAD(f, c, x0::AbstractVector, ncon::Integer)
+  nvar = length(x0)
+  nnzh = nvar * (nvar + 1) / 2
+  nnzj = nvar * ncon
+  return ForwardDiffAD(nnzh, nnzj)
+end
+function ForwardDiffAD(f, x0::AbstractVector)
+  nvar = length(x0)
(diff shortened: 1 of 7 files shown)


src/ad.jl
```diff
@@ -1,7 +1,49 @@
 abstract type ADBackend end
-struct ForwardDiffAD <: ADBackend end
-struct ZygoteAD <: ADBackend end
-struct ReverseDiffAD <: ADBackend end
+struct ForwardDiffAD <: ADBackend
+  nnzh::Int
+  nnzj::Int
+end
+function ForwardDiffAD(f, c, x0::AbstractVector, ncon::Integer)
```
dpo, line 6: Why pass f and c here? How about just nvar and ncon?

guingamp-echecs-stream: This is a slight anticipation for sparse structure and optimization of reversediff. This is in these structures that we will compute and store the "Tape" and everything related to the jacobian/hessien pattern.

tmigot: This is a slight anticipation of sparse structure and optimization of reversediff. The "Tape" and everything related to the jacobian/hessian pattern will be computed (needing `f` and `c`) and stored here.

dpo: Maybe it should be added later then, so that we have functional code at all time and we can clearly see what was added to implement each functionality.

guingamp-echecs-stream: Ok, no problem. I will modify it and rebase.

tmigot: Ok, no problem. I will modify and rebase.

dpo (approved):
Thanks!

abelsiqueira:
Thanks

PR #20, opened 2021-06-11
Parametric NLPModelMeta

tmigot:
Follows [NLPModels.jl PR #353](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/353) with parametric meta.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1,7 +1,7 @@
 export ADNLPModel
 
-mutable struct ADNLPModel <: AbstractNLPModel
-  meta::NLPModelMeta
+mutable struct ADNLPModel{T, S} <: AbstractNLPModel{T, S}
+  meta::NLPModelMeta{T, S}
   counters::Counters
   adbackend::ADBackend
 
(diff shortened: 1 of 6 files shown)


src/nlp.jl
```diff
@@ -46,7 +46,7 @@ function ADNLPModel(
```
abelsiqueira, line 46: It should be `ADNLPModel(f, x0::S; ...)` and `T = eltype(S)`. Unless we can't have other types of `S` for AD

abelsiqueira (changes requested):

src/nlp.jl
```diff
@@ -37,34 +37,36 @@ The following keyword arguments are available to the constructors for constraine
...
 
   return ADNLPModel(meta, Counters(), adbackend, f, x -> T[])
 end
 
 function ADNLPModel(
   f,
-  x0::AbstractVector{T},
+  x0::S,
   lvar::AbstractVector,
   uvar::AbstractVector;
```
abelsiqueira, line 59:
```suggestion
  lvar::S,
  uvar::S;
```
What do think?

tmigot: It looks ok, because this is a constraint in the meta. I just thought it would be more general, in case we modify (again) de types in the meta.

dpo: Abstract types inside structs break type inference.

tmigot: Yes, but this suggestion was in the inputs of the constructor.

dpo: Ah I see, sorry. But how does it work if not all vectors have the same "type"? You have to perform conversions?

tmigot: Right now, if `lvar` has a different type, the error will come from the `NLPModelMeta` constructor instead of the `ADNLPModels` constructor.
My point was just anticipating cases where `lvar` would have a different type than `x0`, e.g. sparse vector. But, it is probably anticipating too much. So, I will update.

abelsiqueira (approved):

abelsiqueira:
Thanks

PR #22, opened 2021-06-30
Add benchmark folder for PR testing [WIP]

tmigot:
@MonssafToukal to help continue the discussion and tests.

PR #23, opened 2021-06-30
Add Jenkins file

tmigot:
Adding this file should be sufficient to test the benchmark in the upcoming PR #22 @MonssafToukal

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,117 @@
+def bmarkFile = 'benchmarks.jl'
+pipeline {
+  agent any
+  environment {
+    REPO_EXISTS = fileExists "$repo"
+  }
+  options {
+    skipDefaultCheckout true
+  }
+  triggers {
+    GenericTrigger(
+     genericVariables: [
+        [
+            key: 'action', 
+            value: '$.action',
+            expressionType: 'JSONPath', //Optional, defaults to JSONPath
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+            expressionType: 'JSONPath', //Optional, defaults to JSONPath
+            regexpFilter: '', //Optional, defaults to empty string
+            defaultValue: '' //Optional, defaults to empty string
+        ],
+        [
+            key: 'org',
+            value: '$.organization.login',
+            expressionType: 'JSONPath', //Optional, defaults to JSONPath
+            regexpFilter: '', //Optional, defaults to empty string
+            defaultValue: 'ProofOfConceptForJuliSmoothOptimizers' //Optional, defaults to empty string
```
dpo, line 32: Change this?

MonssafToukal: It is not necessary. It will only take this value if the variable is undefined. I will have to change the default value in the future.

tmigot: So, it is cleaner to put `''` instead of `'ProofOfConceptForJuliSmoothOptimizers'`?

tmigot: I replaced `'ProofOfConceptForJuliSmoothOptimizers'` by `'JuliaSmoothOptimizers'`

PR #34, opened 2021-09-20
Add `OptimizationProblems` in breakage

tmigot:
Since [PR 66 in OptimizationProblems.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/pull/66), the package uses ADNLP problems.

I also updated the name for DCI.


abelsiqueira:
Thanks

PR #33, opened 2021-09-15
Store tape for gradient

tmigot:

PR #37, opened 2021-10-11
Update tests to use NLPModelsTest 0.5.0

abelsiqueira:


tmigot (approved):
Thank you!
I know NLPModelsIpopt.jl is breaking since #33

PR #38, opened 2021-12-17
Remove `nln` from meta (bump NLPModels 0.18)

tmigot:
rmv `nln` in `nls_meta`


abelsiqueira:
There are failures, otherwise it would be approved.

tmigot:
The error with Julia 1.3 should disapear after #39 , and the error on the nightly is independent of the package (see #39).

dpo (approved):
The errors with nightly seems to be related to ReverseDiff.

PR #39, opened 2021-12-17
Use Julia 1.6 as the new stable version

tmigot:


tmigot:
I think this shows that the error on #38  is independent of `ADNLPModels.jl`.

dpo (approved):

PR #42, opened 2022-02-01
Add docs with current backend

tmigot:
update docstring and examples with the current backend system.

I also added a sentence on `nequ` as I remember students were struggling with it last year.

PR #50, opened 2022-02-11
New constructor for nlp

tmigot:
If that works for you @dpo , I can make the same for `ADNLSModels`.
I will handle the docstring in a separate PR, #42

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -70,16 +70,19 @@ function Hvprod(b::ADBackend, f, x, v)
   return ForwardDiff.derivative(t -> gradient(b, f, x + t * v), 0)
 end
 
-function ForwardDiffAD(nvar::Integer, ncon::Integer, f, x0::AbstractVector)
+function ForwardDiffAD(;
+  nvar::Integer = 0,
+  ncon::Integer = 0,
+  f = x -> sum(x),
+  x0::AbstractVector = rand(0),
+  kwargs...,
+)
+  @assert nvar > 0
+  @lencheck nvar x0
   nnzh = nvar * (nvar + 1) / 2
   nnzj = nvar * ncon
   cfg = ForwardDiff.GradientConfig(f, x0)
   return ForwardDiffAD{typeof(cfg)}(nnzh, nnzj, cfg)
-end
-function ForwardDiffAD(nvar::Integer, f, x0::AbstractVector)
-  nnzh = nvar * (nvar + 1) / 2
-  cfg = ForwardDiff.GradientConfig(f, x0)
-  return ForwardDiffAD{typeof(cfg)}(nnzh, 0, cfg)
 end
 gradient(adbackend::ForwardDiffAD, f, x) = ForwardDiff.gradient(f, x, adbackend.cfg)
 function gradient!(adbackend::ForwardDiffAD, g, f, x)
@@ -99,14 +102,15 @@ end
 
 @init begin
   @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
-    function ZygoteAD(nvar::Integer, ncon::Integer, f, x0::AbstractVector)
+    function ZygoteAD(;
+      nvar::Integer = 0,
+      ncon::Integer = 0,
+      kwargs...,
+    )
+      @assert nvar > 0
       nnzh = nvar * (nvar + 1) / 2
       nnzj = nvar * ncon
       return ZygoteAD(nnzh, nnzj)
-    end
-    function ZygoteAD(nvar::Integer, f, x0::AbstractVector)
-      nnzh = nvar * (nvar + 1) / 2
-      return ZygoteAD(nnzh, 0)
     end
     function gradient(::ZygoteAD, f, x)
       g = Zygote.gradient(f, x)[1]
@@ -120,7 +124,7 @@ end
       return Zygote.jacobian(f, x)[1]
     end
     function hessian(b::ZygoteAD, f, x)
-      return jacobian(ForwardDiffAD(length(x), f, x), x -> gradient(b, f, x), x)
+      return jacobian(ForwardDiffAD(nvar = length(x), f = f, x0 = x), x -> gradient(b, f, x), x)
     end
     function Jprod(::ZygoteAD, f, x, v)
       return vec(Zygote.jacobian(t -> f(x + t * v), 0)[1])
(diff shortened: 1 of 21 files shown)


src/nlp.jl
```diff
@@ -35,11 +35,47 @@ The following keyword arguments are available to the constructors for constraine
...
+end
+
+function ADNLPModel(f, x0::S, lvar::S, uvar::S, c, lcon::S, ucon::S; kwargs...) where {S}
+  return ADNLPModel(ForwardDiffAD, f, x0, lvar, uvar, c, lcon, ucon; kwargs...)
+end
+
+function ADNLPModel(::Type{AD}, f, x0::S, lvar::S, uvar::S, c, lcon::S, ucon::S; kwargs...) where {AD, S}
+  backend = AD(length(x0), length(lcon), f, x0)
+  return ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon, backend; kwargs...)
+end
```
dpo, line 72: That looks good, but could probably use some `@lencheck`s?!

test/nlp/basic.jl
```diff
@@ -29,16 +29,16 @@ function test_autodiff_model()
   for adbackend in (:ForwardDiffAD, :ZygoteAD, :ReverseDiffAD)
     x0 = zeros(2)
     f(x) = dot(x, x)
-    nlp = ADNLPModel(f, x0, adbackend = eval(adbackend)(length(x0), 0, f, x0))
+    nlp = ADNLPModel(f, x0, eval(adbackend)(length(x0), 0, f, x0))
```
dpo, line 32: But didn't you just make it so we could write `ADNLPModel(adbackend, f, x0)`? You're still passing `f` in twice.

src/nlp.jl
tmigot: Thanks for the comments.
It seems a bit redundant as there will be a @lencheck in the `ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon, backend; kwargs...)` call.

tmigot:
I just realized that
```
function ADNLPModel(backend, f, x0::S, lvar::S, uvar::S, c, lcon::S, ucon::S; kwargs...) where {S}
```
is not very convenient for `OptimizationProblems.jl` 

https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/761b0b1e283f0c907b8293e9ba5ea777cbd14fe3/src/ADNLPProblems/arglina.jl#L11

and it is preferable to have a keyword argument for this instead.

dpo:
A keyword for the AD backend? That sounds good to me.

src/nlp.jl
dpo: ok

test/nlp/basic.jl
```diff
@@ -48,24 +48,24 @@ function test_autodiff_model()
...
+      nlp = ADNLPModel(f, x0, unc_adbackend)
+      nlp = ADNLPModel(f, x0, lvar, uvar, unc_adbackend)
+      nlp = ADNLPModel(f, x0, c, lcon, ucon, con_adbackend)
+      nlp = ADNLPModel(f, x0, c, lcon, ucon, con_adbackend, y0 = y0)
+      nlp = ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon, con_adbackend)
+      nlp = ADNLPModel(f, x0, lvar, uvar, c, lcon, ucon, con_adbackend, y0 = y0)
+      @test_throws DimensionError ADNLPModel(f, x0, badlvar, uvar, unc_adbackend)
+      @test_throws DimensionError ADNLPModel(f, x0, lvar, baduvar, unc_adbackend)
+      @test_throws DimensionError ADNLPModel(f, x0, c, badlcon, ucon, con_adbackend)
+      @test_throws DimensionError ADNLPModel(f, x0, c, lcon, baducon, con_adbackend)
```
dpo, line 60: I'm a bit confused here. Isn't the backend supposed to be a kwarg?

tmigot: You have the two options, either provide directly the backend this way, or the keyword argument.
This one is also necessary, if someone make a new backend with different constructor arguments than the existing ones.

tmigot: Do you mean confusing for the unit tests or in the code in general?

If we have just 
```
ADNLPModel(f, x0, badlvar, uvar, unc_adbackend)
```
then we call twice the objective function, etc. which is what we want to avoid.
If we have just the kwargs
```
ADNLPModel(f, x0, badlvar, uvar, backend::Type{AD})
```
then it forces the backend (even user-defined ones) to have the exact same constructors (right now it is `AD(length(x0), length(lcon), f, x0)`).

abelsiqueira: We could make every backend have a single constructor with no positional arguments and arbitrary keyword arguments, like `AD(;kwargs...)`,so the constructor ignores the other arguments, and we can specify them. Like
```
function MyAD(; x0::Vector = zeros(0), kwargs...)
end
```

dpo: Yes, or `MyAD(args...; kwargs...)`.

src/ad.jl
```diff
@@ -70,17 +70,20 @@ function Hvprod(b::ADBackend, f, x, v)
   return ForwardDiff.derivative(t -> gradient(b, f, x + t * v), 0)
 end
 
-function ForwardDiffAD(nvar::Integer, ncon::Integer, f, x0::AbstractVector)
+function ForwardDiffAD(;
+  nvar::Integer = 0,
+  ncon::Integer = 0,
+  f = x -> sum(x),
+  x0::AbstractVector = rand(0),
```
dpo, line 77:
```suggestion
  x0::AbstractVector = rand(nvar),
```

src/ad.jl
```diff
@@ -70,17 +70,20 @@ function Hvprod(b::ADBackend, f, x, v)
   return ForwardDiff.derivative(t -> gradient(b, f, x + t * v), 0)
 end
 
-function ForwardDiffAD(nvar::Integer, ncon::Integer, f, x0::AbstractVector)
+function ForwardDiffAD(;
+  nvar::Integer = 0,
+  ncon::Integer = 0,
+  f = x -> sum(x),
```
dpo, line 76: I find it strange that `f` is a kwarg here. Is it possible to make it a positional arg?

tmigot: I moved `nvar`, `f` and `ncon` as positional arguments in that order.

dpo:
I am so lost on this PR. Could you provide a summary?

tmigot:
The idea is to change the current constructor
```
nlp = ADNLPModel(f, x0, adbackend = ForwardDiffAD(length(x0), f, x0))
```
to
```
nlp = ADNLPModel(f, x0, backend = ForwardDiffAD)
```
so we don't call the function `f` twice. All the changes in the PR are following this principle.

We changed the constructor of the backend to use positional args for:
```
function ForwardDiffAD(
  nvar::Integer = 0,
  f = x -> sum(x),
  ncon::Integer = 0;
  x0::AbstractVector = rand(nvar),
  kwargs...,
)
```

tmigot:
@dpo Following our discussion, I modified the backend constructors so that `nvar` and `f` no longer have default values.

dpo (approved):
Thank you! Looks great now!

PR #51, opened 2022-02-18
Add new constructor for nls problems

tmigot:
Follow-up of #50 
We will fix the `doctstring` in a separate PR.


tmigot:
Closed by #50

PR #41, opened 2021-12-22
New backend organization

tmigot:

PR #55, opened 2022-03-30
New backend organization

tmigot:
This is along the line of #41 .
This is a very large change, so I prepare some pre-docs.

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -1,3 +1,37 @@
+struct ADModelBackend{GB,HvB,JvB,JtvB,JB,HB,GHJ}
+  gradient_backend::GB
+  hprod_backend::HvB
+  jprod_backend::JvB
+  jtprod_backend::JtvB
+  jacobian_backend::JB
+  hessian_backend::HB
+  ghjvprod_backend::GHJ
+end
+
+function ADModelBackend(
+  nvar::Integer,
+  f,
+  ncon::Integer = 0;
+  gradient_backend::Type{GB} = ForwardDiffADGradient,
+  hprod_backend::Type{HvB} = ForwardDiffADHvprod,
+  jprod_backend::Type{JvB} = ForwardDiffADJprod,
+  jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
+  jacobian_backend::Type{JB} = ForwardDiffADJacobian,
+  hessian_backend::Type{HB} = ForwardDiffADHessian,
+  ghjvprod_backend::Type{GHJ} = ForwardDiffADGHjvprod,
+  kwargs...,
+) where {GB,HvB,JvB,JtvB,JB,HB,GHJ}
+  return ADModelBackend(
+    GB(nvar, f, ncon; kwargs...),
+    HvB(nvar, f, ncon; kwargs...),
+    JvB(nvar, f, ncon; kwargs...),
+    JtvB(nvar, f, ncon; kwargs...),
+    JB(nvar, f, ncon; kwargs...),
+    HB(nvar, f, ncon; kwargs...),
+    GHJ(nvar, f, ncon; kwargs...),
+  )
+end
+
 abstract type ADBackend end
 
 throw_error(b) =
@@ -8,6 +42,8 @@ jacobian(b::ADBackend, ::Any, ::Any) = throw_error(b)
 hessian(b::ADBackend, ::Any, ::Any) = throw_error(b)
 Jprod(b::ADBackend, ::Any, ::Any, ::Any) = throw_error(b)
 Jtprod(b::ADBackend, ::Any, ::Any, ::Any) = throw_error(b)
+Hvprod(b::ADBackend, ::Any, ::Any, ::Any) = throw_error(b)
+directional_second_derivative(::ADBackend, ::Any, ::Any, ::Any, ::Any) = throw_error(b)
 function hess_structure!(
   b::ADBackend,
   nlp,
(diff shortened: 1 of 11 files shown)


tmigot:
`ADNLPModels` allows the use of different backends to compute the derivatives required within NLPModel API.
It uses `ForwardDiff.jl`, and via optional depencies `ReverseDiff.jl` and `Zygote.jl`.
I am proposing here a new organization of the backends.

Sum up of the changes per file:
- `ad.jl`: add a new structure `ADModelBackend` ; remove default implementation for `directional_second_derivative` and `Hvprod`.
- `nlp.jl` and `nls.jl`: the `adbackend` attribute is now a `ADModelBackend` ; use `ADModelBackend` attributes for each function in the API.
- `forward.jl`, `reverse.jl` and `zygote.jl`: split the backend in several backends, one per different function. The signature of each constructor is the same.
- In folder `test` we replaced the keyword `backend` by a set of `kwargs`.

The backend information is now in a structure `ADModelBackend` in the attribute `adbackend` of any `ADNLPModel`. 
It acts as a dictionary precising which backend to use for each function.
The functions currently used to define the NLPModel API and the possible backends are defined in the following table (this is not modified by the refactorization):

| Functions | FowardDiff backends | ReverseDiff backends | Zygote backends |
| ----------- | ----------- | ----------- | ----------- |
| `gradient` and `gradient!` | `ForwardDiffADGradient` | `ReverseDiffADGradient` | `ZygoteADGradient` |
| `jacobian` | `ForwardDiffADJacobian` | `ReverseDiffADJacobian` | `ZygoteADJacobian` |
| `hessian` | `ForwardDiffADHessian` | `ReverseDiffADHessian` | `ZygoteADHessian` |
| `Jprod` | `ForwardDiffADJprod` | `ReverseDiffADJprod` | `ZygoteADJprod` |
| `Jtprod` | `ForwardDiffADJtprod` | `ReverseDiffADJtprod` | `ZygoteADJtprod` |
| `Hvprod` | `ForwardDiffADHvprod` | `ReverseDiffADHvprod` | -- |
| `directional_second_derivative` | `ForwardDiffADGHjvprod` | -- | -- |

The functions `hess_structure!`, `hess_coord!`, `jac_structure!` and `jac_coord!` defined in `ad.jl` are generic to all the backends for now.

### Comparison with the current `ADNLPModels.jl`:
In the current verison, there is one structure per AD package, while the suggested version would get one structure per AD package and per operation.

### Comparison with `https://github.com/dpo/ReverseADNLSModels`:
- the backends for each function is controled by a `Dict` ; here it is a structure `ADModelBackend`.
- for each operation the backends are specific here, while in `ReverseADNLSModels` there is a structure per AD package storing all information.

I believe the proposed version allows for a more plug-and-play approach.
This refactorization offers the possibility to quickly use a different backend implementation for one of the function.
Indeed, it suffices to give the desired type with the kwargs when creating the `ADNLPModel`:
  - gradient_backend::Type{GB} = ForwardDiffADGradient,
  - hprod_backend::Type{HvB} = ForwardDiffADHvprod,
  - jprod_backend::Type{JvB} = ForwardDiffADJprod,
  - jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
  - jacobian_backend::Type{JB} = ForwardDiffADJacobian,
  - hessian_backend::Type{HB} = ForwardDiffADHessian,
  - ghjvprod_backend::Type{GHJ} = ForwardDiffADGHjvprod,
The `ADNLPModel` instantiates the `ADModelBackend` structure with the above backends.

The next step would be to optimize the implementation of each of the functions in the table above.
The proposed approach allows to easily benchmark two implementations of a function.

## Examples

Note that the refactorization is non-breaking in the sense that it doesn't change the naive behavior of ADLNPModel.
```
using ReverseDiff, ADNLPModels, NLPModels
nlp = ADNLPModel(sum, ones(3))
grad(nlp, nlp.meta.x0)
```
Thanks to the splitted backends, we can easily change the backend without recreating an `ReverseDiffAD{T} <: ADBackend` as it is now.
```
using ReverseDiff, ADNLPModels, NLPModels
nlp = ADNLPModel(sum, ones(3), gradient_backend = ADNLPModels.ReverseDiffADGradient)
grad(nlp, nlp.meta.x0)
```
We can then rather easily try out some new implementation for each function.
```
using ADNLPModels, NLPModels
struct NewADGradient <: ADNLPModels.ADBackend end
function NewADGradient(
  nvar::Integer,
  f,
  ncon::Integer = 0;
  kwargs...,
)
  return NewADGradient()
end
ADNLPModels.gradient(adbackend::NewADGradient, f, x) = rand(Float64, size(x))
function ADNLPModels.gradient!(adbackend::NewADGradient, g, f, x)
  g .= rand(Float64, size(x))
  return g
end
nlp = ADNLPModel(sum, ones(3), gradient_backend = NewADGradient)
grad(nlp, nlp.meta.x0)
```

dpo:
Is it possible to change, say, the grad backend *after* the model has been created?

tmigot:
> 

It is possible by reassigning the `adbackend`. For instance,
```
adback = ADNLPModels.ADModelBackend(nlp.meta.nvar, nlp.f, gradient_backend = ADNLPModels.ForwardDiffADGradient)
nlp.adbackend = adback
```

dpo:
It would be useful to define methods to reassign the backend in that way so the user doesn't need to pass the other arguments.

tmigot:
Good idea! I add this to the todo-list.

tmigot:
As discussed monday, I move forward with this one.

PR #58, opened 2022-04-14
set `minimize` as ` kwargs`

tmigot:


tmigot:
Discussed during our meeting.

PR #64, opened 2022-04-29
Add setter for backend

tmigot:
As mentioned in https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/55#issuecomment-1109299069

I think we can also close https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/46 for instance:
```
using ADNLPModels, NLPModels
nlp = ADNLPModel(sum, ones(3))
grad(nlp, nlp.meta.x0)
set_adbackend!(nlp, gradient_backend = ADNLPModels.ForwardDiffADGradient, x0 = ones(Float32, 3))
grad(nlp, rand(Float32, 3))
```

### DIFF
--- a/src/ADNLPModels.jl
+++ b/src/ADNLPModels.jl
@@ -15,4 +15,44 @@ include("zygote.jl")
 include("nlp.jl")
 include("nls.jl")
 
+export get_adbackend, set_adbackend!
+
+"""
+    has_adbackend(nlp)
+
+Returns the value `adbackend` from nlp.
+"""
+get_adbackend(nlp::Union{ADNLPModel, ADNLSModel}) = nlp.adbackend
+
+"""
+    set_adbackend!(nlp, new_adbackend)
+    set_adbackend!(nlp; kwargs...)
+
+Replace the current `adbackend` value of nlp by `new_adbackend` or instantiate a new one with `kwargs`, see [`ADModelBackend`](@ref).
+By default, the setter with kwargs will reuse existing backends.
+"""
+function set_adbackend!(nlp::Union{ADNLPModel, ADNLSModel}, new_adbackend::ADModelBackend)
+  nlp.adbackend = new_adbackend
+  return nlp
+end
+function set_adbackend!(nlp::Union{ADNLPModel, ADNLSModel}; kwargs...)
+  args = []
+  for field in fieldnames(ADNLPModels.ADModelBackend)
+    push!(args, if field in keys(kwargs) && typeof(kwargs[field]) <: ADBackend
+      kwargs[field]
+    elseif field in keys(kwargs) && typeof(kwargs[field]) <: DataType
+      if typeof(nlp) <: ADNLPModel
+        kwargs[field](nlp.meta.nvar, nlp.f, nlp.meta.ncon; kwargs...)
+      elseif typeof(nlp) <: ADNLSModel
+        kwargs[field](nlp.meta.nvar, x -> sum(nlp.F(x) .^ 2), nlp.meta.ncon; kwargs...)
+      end
+    else
+      getfield(nlp.adbackend, field)
+    end
+    )
+  end 
+  nlp.adbackend = ADModelBackend(args...)
+  return nlp
+end
+
 end # module
--- a/src/ADNLPModels.jl
+++ b/src/ADNLPModels.jl
@@ -28,7 +28,7 @@ get_adbackend(nlp::Union{ADNLPModel, ADNLSModel}) = nlp.adbackend
     set_adbackend!(nlp, new_adbackend)
     set_adbackend!(nlp; kwargs...)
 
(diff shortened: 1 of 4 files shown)


src/ADNLPModels.jl
```diff
@@ -15,4 +15,44 @@ include("zygote.jl")
 include("nlp.jl")
 include("nls.jl")
 
+export get_adbackend, set_adbackend!
+
+"""
+    has_adbackend(nlp)
```
dpo, line 21:
```suggestion
    get_adbackend(nlp)
```
```suggestion
    has_adbackend(nlp)
```

dpo, line 21:
```suggestion
    get_adbackend(nlp)
```
```suggestion
    has_adbackend(nlp)
```

src/ADNLPModels.jl
```diff
@@ -15,4 +15,44 @@ include("zygote.jl")
...
+        kwargs[field](nlp.meta.nvar, x -> sum(nlp.F(x) .^ 2), nlp.meta.ncon; kwargs...)
+      end
+    else
+      getfield(nlp.adbackend, field)
+    end
+    )
+  end 
+  nlp.adbackend = ADModelBackend(args...)
+  return nlp
+end
```
dpo, line 56: The second method is bit complicated. Does it have an advantage compared to
```
adbackend = ADBackend(...)
set_adbackend!(nlp, adbackend)
```

dpo, line 56: The second method is bit complicated. Does it have an advantage compared to
```
adbackend = ADBackend(...)
set_adbackend!(nlp, adbackend)
```

test/runtests.jl
```diff
@@ -29,6 +29,32 @@ ZygoteAD() = ADNLPModels.ADModelBackend(
   hessian_backend = ADNLPModels.ZygoteADHessian,
 )
 
+function test_getter_setter(nlp)
+  @test get_adbackend(nlp) == nlp.adbackend
+  if typeof(nlp) <: ADNLPModel
+    set_adbackend!(nlp, ReverseDiffAD(nlp.meta.nvar, nlp.f))
```
dpo, line 35: `x -> obj(nlp, x)`?

dpo, line 35: `x -> obj(nlp, x)`?

test/runtests.jl
```diff
@@ -29,6 +29,32 @@ ZygoteAD() = ADNLPModels.ADModelBackend(
...
+  elseif typeof(nlp) <: ADNLSModel
+    set_adbackend!(nlp, ReverseDiffAD(nlp.meta.nvar, x -> sum(nlp.F(x) .^ 2)))
```
dpo, line 37: same here?

dpo, line 37: same here?

dpo:
Sorry, I'm not sure why my comments appear several times...

test/runtests.jl
tmigot: The drawback of using `obj` (same for `residual`) is that every time the function will be used within the AD machinery it will trigger the objective function counter.

For instance, when calling `grad` the AD-system will use the function `obj` and we will get 1 eval of gradient and several evals of obj, and I don't think that's the expected behavior of the counters.

src/ADNLPModels.jl
tmigot: It is more complicated for us, but it is more user-friendly I thought.
There is no advantage apart from not having to know about the `ADBackend` structure.

src/ADNLPModels.jl
tmigot: Thanks for catching that!

tmigot:
Argh, now the tests fail because of the update of `NLPModelsTest` and `NLPModelsModifiers` ...

src/ADNLPModels.jl
dpo: You still have to know all the kwargs. It's basically the same as writing `set_backend!(nlp, ADBackend(...))`, so I don't think there's a significant advantage for the user either, and the implementation seems difficult to maintain.

test/runtests.jl
dpo: I see. That makes sense. Thanks for explaining!

src/ADNLPModels.jl
dpo: I would remove the second method. The rest looks great!

tmigot: Oh sorry, re-reading the docstring I now remember. The positive thing about the second one is that it reuses the existing backends. So, we potentially avoid some computations. But, we can remove it for now and save it for later if needed.

I will remove and merge once the test are fixed.

tmigot:
This was waiting for the unit tests to pass, so I will merge.

PR #66, opened 2022-05-25
Add linear api

tmigot:
Finally, the update with the linear API. Again this is a large PR so here is a recap':
- I updated all the problems in test folder to match the linear API.
- We now have `SparseArrays` in the deps
- `ADNLPModel` and `ADNLSModel` store four vectors for the linear constraint function `Ax + blin`
- There are two new constructor per model + two extras when the user passes a matrix instead of the COO-triplet
- I removed the possibility to provide linear constraints via the `lin` keyword.
- The rest is the implementation of the linear API (recall that for efficiency we also reimplement `jtprod!`).

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -356,40 +356,69 @@ function NLPModels.grad!(nlp::ADNLPModel, x::AbstractVector, g::AbstractVector)
   return g
 end
 
-function NLPModels.cons!(nlp::ADNLPModel, x::AbstractVector, c::AbstractVector)
+function NLPModels.cons_lin!(nlp::ADNLPModel, x::AbstractVector, c::AbstractVector)
   @lencheck nlp.meta.nvar x
-  @lencheck nlp.meta.ncon c
-  increment!(nlp, :neval_cons)
+  @lencheck nlp.meta.nlin c
+  increment!(nlp, :neval_cons_lin)
+  coo_prod!(nlp.clinrows, nlp.clincols, nlp.clinvals, x, c)
+  return c
+end
+
+function NLPModels.cons_nln!(nlp::ADNLPModel, x::AbstractVector, c::AbstractVector)
+  @lencheck nlp.meta.nvar x
+  @lencheck nlp.meta.nnln c
+  increment!(nlp, :neval_cons_nln)
   c .= nlp.c(x)
   return c
 end
 
-function NLPModels.jac(nlp::ADNLPModel, x::AbstractVector)
-  @lencheck nlp.meta.nvar x
-  increment!(nlp, :neval_jac)
-  return jacobian(nlp.adbackend.jacobian_backend, nlp.c, x)
-end
-
-function NLPModels.jac_structure!(
+function NLPModels.jac_lin_structure!(
   nlp::ADNLPModel,
   rows::AbstractVector{<:Integer},
   cols::AbstractVector{<:Integer},
 )
-  @lencheck nlp.meta.nnzj rows cols
+  @lencheck nlp.meta.lin_nnzj rows cols
+  rows .= nlp.clinrows
+  cols .= nlp.clincols
+  return rows, cols
+end
+
+function NLPModels.jac_nln_structure!(
+  nlp::ADNLPModel,
+  rows::AbstractVector{<:Integer},
+  cols::AbstractVector{<:Integer},
+)
+  @lencheck nlp.meta.nln_nnzj rows cols
   return jac_structure!(nlp.adbackend.jacobian_backend, nlp, rows, cols)
 end
 
-function NLPModels.jac_coord!(nlp::ADNLPModel, x::AbstractVector, vals::AbstractVector)
+function NLPModels.jac_lin_coord!(nlp::ADNLPModel, x::AbstractVector, vals::AbstractVector)
   @lencheck nlp.meta.nvar x
-  @lencheck nlp.meta.nnzj vals
-  increment!(nlp, :neval_jac)
+  @lencheck nlp.meta.lin_nnzj vals
+  increment!(nlp, :neval_jac_lin)
+  vals .= nlp.clinvals
+  return vals
+end
+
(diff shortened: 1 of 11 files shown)


src/nlp.jl
```diff
@@ -157,6 +171,82 @@ function ADNLPModel(
   return ADNLPModel(meta, Counters(), adbackend, f, c)
 end
 
+function ADNLPModel(f, x0::S, clinrows, clincols, clinvals::S, blin::S, lcon::S, ucon::S; kwargs...,) where {S}
+  T = eltype(S)
+  return ADNLPModel(f, x0, clinrows, clincols, clinvals, blin, x -> T[], lcon, ucon; kwargs...)
+end
+
+function ADNLPModel(f, x0::S, A::AbstractSparseMatrix{Tv,Ti}, blin::S, lcon::S, ucon::S; kwargs...,) where {S, Tv,Ti}
```
dpo, line 179: The size of A and blin should match.

src/nlp.jl
```diff
@@ -157,6 +171,82 @@ function ADNLPModel(
...
+  clinrows, clincols, clinvals = findnz(A)
+  return ADNLPModel(f, x0, clinrows, clincols, clinvals, blin, c, lcon, ucon; kwargs...)
+end
+
+function ADNLPModel(f, lvar::S, uvar::S, x0::S, clinrows, clincols, clinvals::S, blin::S, lcon::S, ucon::S; kwargs...,) where {S}
+  T = eltype(S)
+  return ADNLPModel(f, x0, lvar, uvar, clinrows, clincols, clinvals, blin, x -> T[], lcon, ucon; kwargs...)
+end
+
+function ADNLPModel(f, x0::S, lvar::S, uvar::S, A::AbstractSparseMatrix{Tv,Ti}, blin::S, lcon::S, ucon::S; kwargs...,) where {S, Tv,Ti}
```
dpo, line 245: Above, you have to give lvar, uvar, x0, but here you have to give x0, lvar, uvar.

src/nlp.jl
```diff
@@ -201,6 +289,66 @@ function ADNLPModel(
   return ADNLPModel(meta, Counters(), adbackend, f, c)
 end
 
+function ADNLPModel(
```
dpo, line 292: Redundant with the constructor on line 240?

src/nlp.jl
```diff
@@ -201,6 +289,66 @@ function ADNLPModel(
...
+    minimize = minimize,
+    islp = false,
+    name = name,
+  )
+  adbackend = ADModelBackend(nvar, f, ncon; x0 = x0, kwargs...)
+
+  return ADNLPModel(meta, Counters(), adbackend, f, c, clinrows, clincols, clinvals, blin)
+end
+
+function ADNLPModel(f, x0, lvar, uvar, A::AbstractSparseMatrix{Tv,Ti}, blin, c, lcon, ucon; kwargs...) where {Tv, Ti}
```
dpo, line 347: Redundant with the constructor on line 245?

src/nlp.jl
```diff
@@ -201,6 +289,66 @@ function ADNLPModel(
...
+  clinrows, clincols, clinvals = findnz(A)
+  return ADNLPModel(f, x0, lvar, uvar, clinrows, clincols, clinvals, blin, c, lcon, ucon; kwargs...)
+end
+
 function NLPModels.obj(nlp::ADNLPModel, x::AbstractVector)
   @lencheck nlp.meta.nvar x
   increment!(nlp, :neval_obj)
```
dpo, line 354: It seems `minimize` is not used. Is that intentional?

src/nlp.jl
```diff
@@ -214,40 +362,82 @@ function NLPModels.grad!(nlp::ADNLPModel, x::AbstractVector, g::AbstractVector)
...
 
-function NLPModels.cons!(nlp::ADNLPModel, x::AbstractVector, c::AbstractVector)
+function NLPModels.cons_lin!(nlp::ADNLPModel, x::AbstractVector, c::AbstractVector{T}) where {T}
+  @lencheck nlp.meta.nvar x
+  @lencheck nlp.meta.nlin c
+  increment!(nlp, :neval_cons_lin)
+  fill!(c, zero(T))
+  for i=1:nlp.meta.lin_nnzj
+    c[nlp.clinrows[i]] += nlp.clinvals[i] * x[nlp.clincols[i]]
+  end
```
dpo, line 372: Here you could call `jprod_lin` to avoid repeated code (or factor out the 3 lines that compute the product to avoid incrementing the wrong counter).

src/nlp.jl
```diff
@@ -256,11 +446,48 @@ function NLPModels.jtprod!(
   nlp::ADNLPModel,
   x::AbstractVector,
   v::AbstractVector,
-  Jtv::AbstractVector,
-)
+  Jtv::AbstractVector{T},
+) where {T}
```
dpo, line 450: Why doesn't this just call `jtprod_lin` and `jtprod_nln`?

dpo (commented):
First round. I only looked at nlp.

src/nlp.jl
tmigot: It is. `minimize` is only informative.

src/nlp.jl
tmigot: Since, we would have to do the sum of the two calls, it will allocate. https://github.com/JuliaSmoothOptimizers/NLPModels.jl/issues/384

tmigot:
@dpo I rebased after the NLS constructors, so it is ready to be reviewed. This PR now includes:
- changing the constructors in the test problems
- implementing the API

Please let me know, if you want to split it further. After this one, we are back on track with the unit tests.

amontoison (approved):

amontoison:
Sorry, I forgot that you have a bot for this package :open_mouth:

amontoison:
Should I revert the PR ?

dpo:
Yes, it cuts both ways that bot...

Thanks for reviewing. We need to move forward with this!

tmigot:
Thanks guys, for supporting this long-term project!

PR #69, opened 2022-06-06
Add constructor with linear constraints

tmigot:
This is the first part of #66 
Add constructors for linear API in `ADNLPModel`. There are constructors for:
- linear constraints only
- bounds + linear constraints
- linear and nonlinear constraints
- bounds + linear and nonlinear constraints

For each constructor, there is a shortcut with a sparse matrix `A` instead of the basic COO-constructor.

Tests pass combining: https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/tree/add-linearAPI + https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/tree/up-linear-cons


src/nlp.jl
```diff
@@ -157,6 +157,81 @@ function ADNLPModel(
...
+  x0::S,
+  clinrows,
+  clincols,
+  clinvals::S,
+  c,
+  lcon::S,
+  ucon::S;
+  y0::S = fill!(similar(lcon), zero(eltype(S))),
+  name::String = "Generic",
+  minimize::Bool = true,
```
dpo, line 181: The docstring doesn't mention what kwargs are accepted.

src/nlp.jl
```diff
@@ -157,6 +157,81 @@ function ADNLPModel(
...
+    lin = lin,
+    lin_nnzj = lin_nnzj,
+    nln_nnzj = nln_nnzj,
+    minimize = minimize,
+    islp = false,
+    name = name,
+  )
+  adbackend = ADModelBackend(nvar, f, ncon; x0 = x0, kwargs...)
+
+  return ADNLPModel(meta, Counters(), adbackend, f, c, clinrows, clincols, clinvals)
```
dpo, line 217: Here you pass c followed by A, but in the constructor above, it's the other way around.

tmigot: True, thanks. I switched in the main structure as well to keep the order linear - nonlinear.

dpo (approved):
Looks great!

PR #72, opened 2022-08-02
add linear API in NLS constructor

tmigot:
This is the NLS-version of https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/69
Again this is extracted from the large PR https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/66 (so I refer to #66 for the unit tests)


src/nls.jl
```diff
@@ -19,15 +26,24 @@ ADNLPModels.show_header(io::IO, nls::ADNLSModel) = println(
...
+    ADNLSModel(F, x0, nequ, lvar, uvar, A, c, lcon, ucon)
 
 ADNLSModel is an Nonlinear Least Squares model using automatic differentiation to
 compute the derivatives.
 The problem is defined as
 
      min  ½‖F(x)‖²
-    s.to  lcon ≤ c(x) ≤ ucon
+    s.to  lcon ≤ (  Ax  ) ≤ ucon
+                 ( c(x) )
```
dpo, line 46: Maybe just to make it clear, state that the linear constraints comes first in the list of constraints.

And so, do you still need the keyword `linequ` below?

dpo: On second reading, maybe `linequ` is to indicate linear *residuals*. If so, that could be explained more clearly.

dpo (approved):

dpo:
Excellent, thank you!

src/nls.jl
tmigot: the linear part of the residual is a bit work to be done, so I will had to the todo-list this clarification. Thanks

PR #82, opened 2022-10-01
add in-place `Hprod!`

tmigot:


paraynaud:
Allocations happened during:
```julia
ForwardDiff.derivative!(Hv, (y, t) -> ForwardDiff.gradient!(y, f, x + t * v), b.gx, 0)
```
especially because of `x + t * v` where there is a `copy`, because `x + t * v` is not inplace and the result is not only a `Vector{T}` but a `::Vector{ForwardDiff.Dual{...}}`.
I replaced `x + t*v` by `x` (knowing there wille be a wrong the result), but it divided by two the amount of allocations.

I hope it helps understanding where allocations are, but I don't have any suggestion on how to make `x + t*v` allocation-free.
I tried `LinearAlgebra.axpby!(1,x,t,v)` without success since it doesn't support `t::ForwardDiff.Dual{...}`.

PR #83, opened 2022-10-12
Add sparse jacobian

tmigot:
I will split this into several PRs as it is a large change

- [x] Add `c` in backend constructors
- [x] Add `get_nln_nnzj` and `get_nln_nnzh` functions
- [x] Extend ADModelBackend to NLS functions
- [ ] Add sparse jacobian!


tmigot:
I also added a complementary tests with problems from OptimizationProblems.jl:
```
using ADNLPModels, LinearAlgebra, NLPModels, NLPModelsJuMP, OptimizationProblems

for prob in intersect(names(ADNLPProblems), names(PureJuMP))
  if prob == :hs61
    continue
  end
  if occursin("tetra", string(prob)) || occursin("triangle", string(prob))
    continue # too slow
  end
  nlp = eval(Meta.parse("ADNLPProblems.$(prob)()"))
  model = MathOptNLPModel(eval(Meta.parse("PureJuMP.$(prob)()")))
  if nlp.meta.ncon > 0
    x = nlp.meta.x0
    test = Matrix(jac(nlp, x)) ≈ Matrix(jac(model, x))
    if !test
      println("$prob : $(nlp.meta.nvar), $(nlp.meta.ncon)")
      @show norm(Matrix(jac(nlp, x)) - Matrix(jac(model, x)))
    end
  end
end
```
and all the tests passed.

src/sparse_derivatives.jl
```diff
@@ -0,0 +1,38 @@
...
+  cfJ = Symbolics.build_function(vals, w, expression = Val{false})
+  SparseADJacobian(nnzj, rows, cols, cfJ)
+end
+
+function get_nln_nnzj(b::SparseADJacobian, nvar, ncon)
+  b.nnzj
+end
+
+function jac_structure!(
+  b::SparseADJacobian,
```
amontoison, line 25: `b` should be stored in the `nlp` model.
The user doesn't want to provide a specific structure for `Jacobian`, `Hessian`, etc...
We will have a reason to `nlp` as an argument.

src/nlp.jl
```diff
@@ -329,6 +330,7 @@ function ADNLPModel(
     lcon = lcon,
     ucon = ucon,
     nnzj = nnzj,
+    nln_nnzj = nnzj,
```
amontoison, line 333: Do we need to provide `nln_nnzj` with the recent API of `NLPModels.jl`?
You don't seem to use it later in the code.

src/sparse_derivatives.jl
```diff
@@ -0,0 +1,38 @@
+struct SparseADJacobian <: ADBackend
+  nnzj
+  rows
+  cols
+  cfJ
```
amontoison, line 5: Add the type of some attributes if you can. It will help the compiler otherwise it will infer the type at runtime.

src/nlp.jl
tmigot: Well, yes. I didn't force `nnzj = nln_nnzj _ lin_nnzj` in the NLPModelMeta https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/a2ce45edc004a6e0adce7c49074619d6a467caa8/src/nlp/meta.jl#L71 . The reason was that some packages/users are not using the linear API.

It is used in `jac_nln_structure!` and `jac_nln_coord!. `nnzj` is not used in the code, but it is used by some solvers.

src/sparse_derivatives.jl
tmigot: It is stored inside `nlp`, see `ADModelBackend` and `ADNLPModel`. We use the `fun(::Backend, ::AbstractNLPModel, args...; kwargs...)` format for the dispatch.

amontoison: In that case, is not possible to define:
```julia
function jac_coord!(nlp, x::AbstractVector, vals::AbstractVector)
  _fun = eval(nlp.b.cfJ[2])
  Base.invokelatest(_fun, vals, x)
  return vals
end
```
?

tmigot: This is done here https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/6e45688a8352de83e184b93fa136de1f2b4aff61/src/nlp.jl#L467

We use this trick to broadcast on the backend as there are more than one implementation.

amontoison: Ok thanks for the clarification, just add `nlp::ADNLPModel` and I will approve the PR.

tmigot: Thanks for the suggestion. Unfortunately, this will require some re-organization of the files (because part of theses functions are defined before ADNLPModel and ADNLSModel).
What about we make an issue **Review typing in backend structures** and fix this in the following PR?

amontoison: Yes, that's fine for me!

amontoison (approved):

dpo:
We really have to update the config of that dratted probot. I didn't even have a chance to look at this PR!

tmigot:
Sorry for that, if you have some feedback on this I can open a new PR to fix it.

https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/91 should help with the bot.

PR #84, opened 2022-10-12
Add constraint function to constructors

tmigot:
Needed for #83


dpo (approved):
Sorry for the delay. For some reason, I missed this one.

PR #86, opened 2022-10-28
Add `get_nnzj` function

tmigot:
Needed for #83

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -78,6 +78,21 @@ end
 
 abstract type ADBackend end
 
+"""
+    get_nln_nnzj(::ADBackend, nvar, ncon)
+    get_nln_nnzj(b::ADModelBackend, nvar, ncon)
+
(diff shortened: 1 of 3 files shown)


src/ad.jl
```diff
@@ -78,6 +78,21 @@ end
 
 abstract type ADBackend end
 
+"""
+    get_nln_nnzj(::ADBackend, nvar, ncon)
+    get_nln_nnzj(b::ADModelBackend, nvar, ncon)
+
+For a given `ADBackend` of a problem with `nvar` variables and `ncon` constraints, it returns the number of nonzeros in the Jacobian of nonlinear constraints.
```
dpo, line 85:
```suggestion
For a given `ADBackend` of a problem with `nvar` variables and `ncon` constraints, return the number of nonzeros in the Jacobian of nonlinear constraints.
```

dpo (approved):
Good idea! Good to merge after this

tmigot:
thanks for the review! I will fix the doc in the next one.

PR #101, opened 2023-02-23
[WIP] Setup in-place constructors

tmigot:


tmigot:
a) The error with DCISolver is the following:
```
f(x) = (x[1] - 1)^2 + 4 * (x[2] - x[1]^2)^2
nlp = ADNLPModel(f, [-1.2; 1.0])
hess(nlp, nlp.meta.x0, nlp.meta.y0)
```
b) The support with Zygote would be broken as it doesn't support broadcast. The only solution would be to pass `ADNLPModel!` with a `c!` that doesn't use broadcast.

PR #110, opened 2023-03-16
Optimize src/ad.jl

amontoison:
@tmigot
- It seems that you forgot to divide by 2 in the definition of `f` for NLS (line 122);
- We should maybe add a `get_f` because we redefine it in many cases;
- ℓ(x) is not defined anymore with a `if ... else ... end`, I think that it could help Symbolics if we remove this kind of "branch".


tmigot (approved):
Thanks for catching the typo !

PR #111, opened 2023-03-16
Add Symbolics and SparseDiffTools Hessian

amontoison:
@tmigot


amontoison:
@tmigot The Hessian requires some modifications compared to the Jacobian. We need to add `obj_weight` and `y` as keyword arguments of `ℓ` such that Symbolics.jl doesn't handle them as constants.

PR #127, opened 2023-04-04
Sparse hessian

amontoison:
@tmigot I have the wrong result when I use `autoauto_color_hessian!`.
I suspect an error in SparseDiffTools.jl. :thinking:

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -74,7 +74,7 @@ function ADModelBackend(
   jprod_backend::Type{JvB} = ForwardDiffADJprod,
   jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
   jacobian_backend::Type{JB} = SparseForwardADJacobian,
-  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseSymbolicsADHessian
+  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseADHessian
   ghjvprod_backend::Type{GHJ} = ForwardDiffADGHjvprod,
   kwargs...,
 ) where {GB, HvB, JvB, JtvB, JB, HB, GHJ}
@@ -105,7 +105,7 @@ function ADModelNLSBackend(
   jprod_backend::Type{JvB} = ForwardDiffADJprod,
   jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
   jacobian_backend::Type{JB} = SparseForwardADJacobian,
-  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseSymbolicsADHessian
+  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseADHessian
   ghjvprod_backend::Type{GHJ} = ForwardDiffADGHjvprod,
   hprod_residual_backend::Type{HvBLS} = ForwardDiffADHvprod,
   jprod_residual_backend::Type{JvBLS} = ForwardDiffADJprod,
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -74,7 +74,7 @@ function ADModelBackend(
   jprod_backend::Type{JvB} = ForwardDiffADJprod,
   jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
   jacobian_backend::Type{JB} = SparseForwardADJacobian,
-  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseADHessian
+  hessian_backend::Type{HB} = SparseADHessian,
   ghjvprod_backend::Type{GHJ} = ForwardDiffADGHjvprod,
   kwargs...,
 ) where {GB, HvB, JvB, JtvB, JB, HB, GHJ}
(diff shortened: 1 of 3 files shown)


dpo:
Could we see a benchmark with JuMP just to know where this stands?

amontoison:
@tmigot Is it possible to do have a benchmark with the `SparseHessian` backend based on Symbolics.jl for Dominique?

tmigot:
The benchmark JuMP vs Symbolics Hessian:
https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/benchmark/benchmark/2023-04-09_adnlpmodels_benchmark_hess_nscal_100_mono.png
for problems with about 100 variables and constraints.

tmigot (changes requested):
Can you also change these two lines:
https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/327bd6373be23901fbac2e87ec7cf142326699f0/src/ad.jl#L77

https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/327bd6373be23901fbac2e87ec7cf142326699f0/src/ad.jl#L108

to have this new backend by default?

tmigot:
2023-04-27_adnlpmodels_benchmark_hess_nscal_1000_mono
Not bad!

src/sparse_hessian.jl
```diff
@@ -1,3 +1,100 @@
+## ----- SparseDiffTools -----
```
tmigot, line 1: Should we remove this comment?

amontoison: Yes, I also propose the rename this backend `SparseADHessian`.

src/ad.jl
```diff
@@ -74,7 +74,7 @@ function ADModelBackend(
   jprod_backend::Type{JvB} = ForwardDiffADJprod,
   jtprod_backend::Type{JtvB} = ForwardDiffADJtprod,
   jacobian_backend::Type{JB} = SparseForwardADJacobian,
-  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseSymbolicsADHessian
+  hessian_backend::Type{HB} = ForwardDiffADHessian, # sparse alternative: SparseADHessian
```
tmigot, line 77:
```suggestion
  hessian_backend::Type{HB} = SparseADHessian,
```

src/ad.jl
```diff
@@ -105,7 +105,7 @@ function ADModelNLSBackend(
...
```
tmigot, line 108:
```suggestion
  hessian_backend::Type{HB} = SparseADHessian,
```

tmigot (changes requested):

amontoison:
> Not bad! Sparse is this one, Forward is the dense ForwardDiff.
> 
> 2023-04-27_adnlpmodels_benchmark_hess_nscal_1000_mono

Nice! Can you add the `Symbolic` version too?
We didn't optimize the directional derivative of the gradient this afternoon so it's already great!

src/sparse_hessian.jl
```diff
@@ -1,3 +1,98 @@
...
+  b::SparseADHessian,
+  nlp::ADModel,
+  x::AbstractVector,
+  obj_weight::Real,
+  vals::AbstractVector,
+)
+  ℓ(x) = obj_weight * nlp.f(x)
+  sparse_hess_coord!(ℓ, b, x, vals)
+end
+
```
tmigot, line 95: The tests should pass after this:
```suggestion

function hess_coord!(
  b::SparseADHessian,
  nlp::ADModel,
  x::AbstractVector,
  j::Integer,
  vals::AbstractVector{T},
) where {T}
  y = zeros(T, nlp.meta.nnln)
  for (w, k) in enumerate(nlp.meta.nln)
    y[w] = k == j ? 1 : 0
  end
  obj_weight = zero(T)
  ℓ = get_lag(nlp, b, obj_weight, y)
  sparse_hess_coord!(ℓ, b, x, vals)
  return vals
end
```

src/sparse_hessian.jl
```diff
@@ -1,3 +1,98 @@
...
+function hess_coord!(
+  b::SparseADHessian,
+  nlp::ADModel,
+  x::AbstractVector,
+  obj_weight::Real,
+  vals::AbstractVector,
+)
+  ℓ(x) = obj_weight * nlp.f(x)
```
tmigot, line 92:
```suggestion
  ℓ = get_lag(nlp, b, obj_weight)
```
as `nlp.f` is not defined for NLS

tmigot (approved):
Looks good to me !
Next steps:
- [ ] improve hprod;
- [ ] add our own algo for thé coloration.

jbcaillau:
@amontoison @tmigot Hi guys, looks good: any release available for testing on sparse Jacobians / Hessians?

amontoison:
Hi @jbcaillau!
Tangi, do you think that we can do a release after #139?

tmigot:
Actually, we just need to try and solve https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/140 first, but next release should be in a week or 2. I will keep you guys updated when it is done.

jbcaillau:
@tmigot @amontoison hi there, any news on this? Actually, can you please update me on what might be pulled to test solving with sparse features?

tmigot:
Hi @jbcaillau ! ADNLPModels 0.6.2 has the sparse Hessian and Jacobian implemented. By the end of the month, there will be another upgraded version with improved `jprod`, `jtprod` and `hprod` and faster `Hessian/jacobian`.

jbcaillau:
Thanks @tmigot for the confirmation. I am forwarding to the ct team that they can rely on 0.6.2 to benchmark sparse derivatives 👍🏽

amontoison:
@jbcaillau 
We released [JuliaHSL](https://licences.stfc.ac.uk/product/julia-hsl) this week.
If you use Ipopt.jl, you will probably see a huge difference.

jbcaillau:
thanks @amontoison : does it require to apply for an academic license?

amontoison:
Yes, you need to apply for an academic license.

jbcaillau:
✅ order passed 🤞🏾

jbcaillau:
Hi @tmigot @amontoison! Not sure to be on the right issue, but anyway: as JuliaCon approaches, we are on the way to announce the last release of [control-toolbox](https://control-toolbox.org). Any news of performance increase for sparse Hessians?

tmigot:
Hi @jbcaillau ! I will make a last attempt on Monday to extend to the Lagrangian what we did there https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/171 . I keep you updated.

PR #139, opened 2023-04-29
Rename SparseForwardADJacobian backend

amontoison:
We will be consistent with the name used for sparse Hessian backends.


tmigot (approved):
Looks great, thanks!

PR #160, opened 2023-06-05
Add hprod with ForwardDiff [WIP]

tmigot:


tmigot:
https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/tree/benchmark/benchmark

PR #230, opened 2024-05-18
Use SparseConnectivityTracer.jl

### LINKED ISSUE #232: Bug with upgrade to SymbolicUtils release 1.6 and 1.7
Hi there; recent problem when calling `Ipopt` on an `ADNLModel`, which is what we do in [OptimalControl.jl](https://control-toolbox.org/docs/optimalcontrol/stable). Was working fine until last week, I would say (first ran into the problem one day ago). Attached is a snapshot of the error, seemingly connected to the sparsity pattern analysis of the Hessian by means of `Symbolics.jl` (and `SymbolicUtils`). The problem can be reproduced by running the basic example from `OptimalControl.jl` [here](https://control-toolbox.org).

IMG_3135
IMG_3136

amontoison:
fix #232 
@gdalle


gdalle:
Can you try it with the branch from https://github.com/adrhill/SparseConnectivityTracer.jl/pull/75 ?

tmigot:
Hi @amontoison @gdalle ! Thanks for adding this new feature. Can you make it that we don't remove the Symbolics-way of computing the sparse matrices? I am fine having both ways.

amontoison:
@tmigot What do you want to keep Symbolics.jl?

tmigot:
> @tmigot What do you want to keep Symbolics.jl?

Why not? I don't think that the debate is 100% clear that the SparseConnectivityTracer is better than Symbolics, and it doesn't cost us a lot to maintain Symbolics for now.

gdalle:
If you want to switch easily between both, I'd suggest using the ADTypes interface for sparsity detection. SparseConnectivityTracer supports it natively, and Symbolics will once https://github.com/JuliaSymbolics/Symbolics.jl/pull/1134 is merged. Then it will be as easy as switching the "detector" object

gdalle:
Wait for https://github.com/adrhill/SparseConnectivityTracer.jl/issues/90

amontoison:
@gdalle Do you have functions to perform `J*u`, `J'*v`, `H*w` in `SCT.jl`?
I already dropped `Symbolics.jl` and I can drop `SparseDiffTools.jl` is I have these routines.

amontoison:
@gdalle It seems that this case is not supported in ADTypes: https://github.com/amontoison/ADNLPModels.jl/blob/SCT/src/sparsity_pattern.jl#L9-L13

gdalle:
> Do you have functions to perform J*u, J'*v, H*w in SCT.jl?

What do you mean by that? The result of a matrix-vector product is dense in general, why would you want its sparsity pattern?

gdalle:
> @gdalle It seems that this case is not supported in ADTypes: https://github.com/amontoison/ADNLPModels.jl/blob/SCT/src/sparsity_pattern.jl#L9-L13

You need to use the function from ADTypes with the detector

```julia
function compute_jacobian_sparsity(c, x0)
  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
  S = ADTypes.jacobian_sparsity(c, x0, detector)
  return S
end
```

amontoison:
> > @gdalle It seems that this case is not supported in ADTypes: https://github.com/amontoison/ADNLPModels.jl/blob/SCT/src/sparsity_pattern.jl#L9-L13
> 
> You need to use the function from ADTypes with the detector
> 
> ```julia
> function compute_jacobian_sparsity(c, x0)
>   detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
>   S = ADTypes.jacobian_sparsity(c, x0, detector)
>   return S
> end
> ```
My bad, I did the typo...

gdalle:
> @gdalle Do you have functions to perform J*u, J'*v, H*w in SCT.jl?

These functions are not connected to the sparsity pattern. The right way to perform them is with DifferentiationInterface

amontoison:
@gdalle Can you review the PR?

docs/src/predefined.md
```diff
@@ -49,30 +49,12 @@ get_adbackend(nlp)
 
 ## Hessian and Jacobian computations
 
-It is to be noted that by default the Jacobian and Hessian matrices are dense.
+It is to be noted that by default the Jacobian and Hessian matrices are sparse.
```
gdalle, line 52: this is a breaking change?

docs/src/predefined.md
```diff
@@ -92,4 +74,6 @@ x = rand(T, 2)
 jac(nlp, x)
 ```
 
-The package [`Symbolics.jl`](https://github.com/JuliaSymbolics/Symbolics.jl) is used to compute the sparsity pattern of the sparse matrix. The evaluation of the number of directional derivatives needed to evaluate the matrix is done by [`ColPack.jl`](https://github.com/michel2323/ColPack.jl).
+The package [`SparseConnectivityTracer.jl`](https://github.com/adrhill/SparseConnectivityTracer.jl) is used to compute the sparsity pattern of Jacobians and Hessians.
+The evaluation of the number of directional derivatives needed to evaluate the matrix is done by [`ColPack.jl`](https://github.com/exanauts/ColPack.jl).
+We acknowledge Guillaume Dalle (@gdalle), Adrian Hill (@adrhill), and Michel Schanen (@michel2323) for the development of these packages.
```
gdalle, line 79: <3

docs/src/predefined.md
```diff
@@ -92,4 +74,6 @@ x = rand(T, 2)
 jac(nlp, x)
 ```
 
-The package [`Symbolics.jl`](https://github.com/JuliaSymbolics/Symbolics.jl) is used to compute the sparsity pattern of the sparse matrix. The evaluation of the number of directional derivatives needed to evaluate the matrix is done by [`ColPack.jl`](https://github.com/michel2323/ColPack.jl).
+The package [`SparseConnectivityTracer.jl`](https://github.com/adrhill/SparseConnectivityTracer.jl) is used to compute the sparsity pattern of Jacobians and Hessians.
+The evaluation of the number of directional derivatives needed to evaluate the matrix is done by [`ColPack.jl`](https://github.com/exanauts/ColPack.jl).
```
gdalle, line 78: strictly speaking it's the evaluation of the seeds (the vectors with which we compute matrix-vector products)

src/sparse_hessian.jl
```diff
@@ -24,8 +24,7 @@ function SparseADHessian(
   kwargs...,
 ) where {S}
   T = eltype(S)
-  Hs = compute_hessian_sparsity(f, nvar, c!, ncon)
-  H = ncon == 0 ? Hs : Hs[1:nvar, 1:nvar]
```
gdalle, line 28: why does this go away?

src/sparse_hessian.jl
```diff
@@ -95,8 +94,7 @@ function SparseReverseADHessian(
   alg = ColPackColoration(),
   kwargs...,
 ) where {T}
-  Hs = compute_hessian_sparsity(f, nvar, c!, ncon)
-  H = ncon == 0 ? Hs : Hs[1:nvar, 1:nvar]
```
gdalle, line 99: why does this go away?

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
+"""
+    compute_jacobian_sparsity(c, x0)
+    compute_jacobian_sparsity(c!, cx, x0)
+
+Return a sparse matrix.
+"""
+function compute_jacobian_sparsity end
+
+function compute_jacobian_sparsity(c, x0)
+  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
```
gdalle, line 10: it would be nice if you could figure out a way to pass the detector all the way down from the user-facing API to the low-level functions

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
...
+
+Return a sparse matrix.
+"""
+function compute_hessian_sparsity(f, nvar, c!, ncon)
+  function lagrangian(x)
+    if ncon == 0
+      return f(x)
+    else
+      cx = zeros(eltype(x), ncon)
+      y0 = rand(ncon)
```
gdalle, line 32: this should prevent accidental cancellations, like we discussed

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
+"""
+    compute_jacobian_sparsity(c, x0)
+    compute_jacobian_sparsity(c!, cx, x0)
+
+Return a sparse matrix.
```
gdalle, line 5: a sparse boolean matrix

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
...
+function compute_jacobian_sparsity(c!, cx, x0)
+  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
+  S = ADTypes.jacobian_sparsity(c!, cx, x0, detector)
+  return S
+end
+
+"""
+    compute_hessian_sparsity(f, nvar, c!, ncon)
+
+Return a sparse matrix.
```
gdalle, line 24: a sparse boolean matrix

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
...
+    if ncon == 0
+      return f(x)
+    else
+      cx = zeros(eltype(x), ncon)
+      y0 = rand(ncon)
+      return f(x) + dot(c!(cx, x), y0)
+    end
+  end
+
+  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
```
gdalle, line 37: same remark about passing the detector

src/sparsity_pattern.jl
```diff
@@ -0,0 +1,41 @@
...
+function compute_jacobian_sparsity end
+
+function compute_jacobian_sparsity(c, x0)
+  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
+  S = ADTypes.jacobian_sparsity(c, x0, detector)
+  return S
+end
+
+function compute_jacobian_sparsity(c!, cx, x0)
+  detector = SparseConnectivityTracer.TracerSparsityDetector()  # replaceable
```
gdalle, line 16: same remark about passing the detector

test/runtests.jl
```diff
@@ -3,21 +3,18 @@ using ADNLPModels, ManualNLPModels, NLPModels, NLPModelsModifiers, NLPModelsTest
 using ADNLPModels:
   gradient, gradient!, jacobian, hessian, Jprod!, Jtprod!, directional_second_derivative, Hvprod!
 
-@testset "Error without loading package for sparsity pattern" begin
+@testset "Test sparsity pattern of Jacobian and Hessian" begin
```
gdalle, line 6: that's a fairly minimal test :rofl:

gdalle (approved):
looks good to me!

src/sparse_hessian.jl
amontoison: Because we don't need anymore to use the Lagrange multipliers as variables. It was mandatory with Symbolics.jl...

docs/src/predefined.md
amontoison: Yes, but we want that by default.
It was not possible before because of the optional dependency on Symbolics.jl.
We didn't wanted to download all SciML packages...

src/sparsity_pattern.jl
amontoison: Done :+1:

tmigot:
Hi guys! Thanks @gdalle and @amontoison for this.
Why did you end up droping Symbolics in the same PR?
Have you guys run some performance benchmark?

gdalle:
There are several reasons, which @amontoison summed up here: https://github.com/control-toolbox/CTDirect.jl/issues/88#issuecomment-2150412031.
A big one was the compatibility issues of Symbolics (some recent versions were buggy, and it hasn't supported Julia 1.6 for a while) and the general weight of the SciML stack of dependencies, as opposed to the very lightweight SparseConnectivityTracer.

We didn't benchmark specifically against Symbolics on optimization problems. However we did run other benchmarks:

- SparseConnectivityTracer against Symbolics on the Jacobian test case used in the article introducing Symbolics sparsity detection: https://github.com/adrhill/SparseConnectivityTracer.jl/pull/4#issuecomment-2093461551 
- SparseConnectivityTracer against JuMP: https://github.com/adrhill/SparseConnectivityTracer.jl/issues/118

Finally, note that SparseConnectivityTracer is only a default choice, not an eternal commitment. Thanks to the new ADTypes sparsity detection API, going back to Symbolics is basically as easy as switching the detector object here:

https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/29b1d2d4d4233041567b66a6645c24c6eba95888/src/sparsity_pattern.jl#L9-L16

I don't think this detector configuration is accessible to the user yet, because it is still in the low-level functions, but it would be good to pass this object down all the way from the top level API

tmigot:
> There are several reasons, which @amontoison summed up here: [control-toolbox/CTDirect.jl#88 (comment)](https://github.com/control-toolbox/CTDirect.jl/issues/88#issuecomment-2150412031). A big one was the compatibility issues of Symbolics (some recent versions were buggy, and it hasn't supported Julia 1.6 for a while)

Ok, but it is not a real constraint, it was just a versioning mistake that is being resolved.

> and the general weight of the SciML stack of dependencies, as opposed to the very lightweight SparseConnectivityTracer.

Agreed.

I am in general very in favor of what you guys did here. My point was mainly that it is strange to do everything in the same PR especially if we don't benchmark this package. The correct strategy should have been:
- add SparseConnectivityTracer
- add a benchmark SparseConnectivityTracer/Symbolics
- discuss about droping Symbolics ? (also it was only an optional dep so it didn't slow down the user experience, only the unit tests)

Could you open an issue to remember us to make the detector configuration accessible to users? Thanks!

gdalle:
I agree with the temporary aspect of the SymbolicUtils kerfuffle. However, Julia 1.6 support was dropped some time ago by Symbolics and it isn't coming back. So that's a real limitation if you want to continue supporting 1.6, at least until 1.10 is declared as the new LTS.
I agree that big changes all at once are not ideal.
Initially I tried a soft transition in PR #238 (just using the detector framework, but sticking with Symbolics). I got the tests working on 1.10 but they failed on 1.6 because of the missing SymbolicsSparsityDetector object (which was introduced after 1.6 support was dropped from Symbolics). I guess the SymbolicsSparsityDetector could have been backported to a version of Symbolics that does support 1.6, but given the tiresomeness of it all, Alexis and I thought it would be best to do a clean break

gdalle:
I opened an issue to keep track of the detector configuration, and I'll add benchmarks against Symbolics to my SCT/JuMP comparison in https://github.com/adrhill/SparseConnectivityTracer.jl/issues/118

amontoison:
> > There are several reasons, which @amontoison summed up here: [control-toolbox/CTDirect.jl#88 (comment)](https://github.com/control-toolbox/CTDirect.jl/issues/88#issuecomment-2150412031). A big one was the compatibility issues of Symbolics (some recent versions were buggy, and it hasn't supported Julia 1.6 for a while)
> 
> Ok, but it is not a real constraint, it was just a versioning mistake that is being resolved.

I wanted to have sparse AD by default here.
The AD was sparse only if Symbolics.jl was installed.
Note thay Symbolics.jl downloads all the internet when we install it.
Because it was an optional dependency, we were unable to set a compat entry on it and depending on multiple factors, the version installed could be incompatible or broken....
It's not something that I want for the users of ADNLPModels.jl.

> > and the general weight of the SciML stack of dependencies, as opposed to the very lightweight SparseConnectivityTracer.
> 
> Agreed.
> 
> I am in general very in favor of what you guys did here. My point was mainly that it is strange to do everything in the same PR especially if we don't benchmark this package. The correct strategy should have been:
> 
> * add SparseConnectivityTracer
> * add a benchmark SparseConnectivityTracer/Symbolics
> * discuss about droping Symbolics ? (also it was only an optional dep so it didn't slow down the user experience, only the unit tests)

I wanted to drop Symbolics.jl and SparsityPatternDetection.jl and it was the best strategy for me.
Even if we lose performances (in the worst cade), we can improve things with Guillaume and Adrian.
We opened issues in Symbolics.jl two years ago and nothing was solved...

> Could you open an issue to remember us to make the detector configuration accessible to users? Thanks!

It's already available as a keyword argument for the sparse AD backends (`SparseADHessian` and `SparseADJacobian`).

PR #231, opened 2024-05-18
[WIP] Transition to package extensions

kiranshila:
Since Julia 1.9, the recommended way to have optional dependencies and additional code that adds support for those optional dependencies is [Package Extensions](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)). This is a WIP PR to reorganize ADNLPModels to support them.

# Current Blockers
- The Symbolics extension introduces new symbols, something that is generally unsupported by extensions. There is a workaround on the dev docs [here](https://pkgdocs.julialang.org/dev/creating-packages/#Behavior-of-extensions), but I can't seem to make it work in tests
- This transition bumps the julia compat up to 1.9, away from the LTS release of 1.6. Is this acceptable? There are workarounds to keep the compat while using package extensions but I thought I'd try to just get extensions working first.


kiranshila:
Actually, I think this is way harder than I expected because of all the new data types. It seems like this might need some serious restructuring to fit into the package extension style.

tmigot:
Hi @kiranshila ! Thanks for working on this. Droping Julia version before 1.9 is a big decision though that I would prefer to delay for the moment if that's ok for you?

```I took the liberty of also removing the test Project.toml as current best practice is to define all the dependencies (including test deps) in the top level toml file and use extras and targets.```

Do you have any link related to this?

Issue #229, opened 2024-05-16
Useful packages for (sparse) differentiation?


gdalle:
Hey there friends of JSO!

@adrhill and I have been hard at work on two autodiff toolkits which you might find useful:

- [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl), an interface to every possible autodiff package with support for derivative, gradient, jacobian, hessian, JVP, VJP, HVP. It also supports sparse autodiff with a coloring-based approach.
- [SparseConnectivityTracer.jl](https://github.com/adrhill/SparseConnectivityTracer.jl), a much faster alternative to Symbolics.jl for sparsity pattern detection.

We would be happy to discuss potential uses, and integration with your ecosystem. We're having similar talks with Optimization.jl and the SciML organization at large.

Ping @amontoison @abelsiqueira since we've already chatted a few times :)

amontoison:
@gdalle
I propose to start by replacing `SparseDiffTools.jl` and `Symbolics.jl` by `SparseConnectivityTracer.jl`.

gdalle:
Sounds good! We're currently in the middle of a big refactor for SparseConnectivityTracer.jl, so it will have to wait a few days

adrhill:
The refactor has been merged and SparseConnectivityTracer `v0.4.0` has just been tagged, so it's ready to be broken by you folks! 😁

PRs adding `@test_broken` tests are more than welcome.

amontoison:
@adrhill @gdalle 
I started to interface `SparseConnectivityTracer.jl` quickly tonight and I have a few feedbacks:
- Is `SparseConnectivityTracer.jacobian_pattern(c!, cx, x0)` more efficient than `SparseConnectivityTracer.jacobian_pattern(c, x0)` ?
Note that `SparseConnectivityTracer.jacobian_pattern(c!, y, x)` is not documented.
- Is it possible to add an option to only return one triangle of the sparsity pattern of the Hessian? You could add a keyword argument with possible values `F` (default), `L` or `U`.
- Can we exploit the symmetry of the hessian to determine it more efficiently?
- In optimization we want the Hessian of the augmented Lagrangian when we have constraints so it means that we want the sparsity structure of the Hessian of `ℓ(x) = f(x) + dot(c(x), y)`. I have an error because `SparseConnectivityTracer` doesn't like `dot`. I can hack it with the help of a `sum`.
[...]

PR #238, opened 2024-06-04
Use generic sparsity detector

### LINKED ISSUE #232: Bug with upgrade to SymbolicUtils release 1.6 and 1.7

gdalle:
ADTypes.jl v1.0 introduced an abstract [sparse autodiff interface](https://sciml.github.io/ADTypes.jl/stable/#Sparse-AD) for sparsity pattern detection and matrix coloring, which can be implemented by various packages. The latest release of Symbolics.jl, namely [v5.29](https://github.com/JuliaSymbolics/Symbolics.jl/releases/tag/v5.29.0), includes a `SymbolicsSparsityDetector` object which satisfies the ADTypes.jl requirements of sparsity pattern detection.

My PR changes your sparsity detection routines to use this detector object, in order to make it easier to switch later on. For instance, you may want to use `TracerSparsityDetector` from SparseConnectivityTracer.jl instead (see #229 and #230).
The only caveat is that it uses a random vector of Lagrange multiplier, but I think that's correct since the probability of accidental cancellation between constraints is basically zero?

I also removed the compat tweak on SymbolicUtils.jl in the test environment, since it has released a v2.0 which apparently fixes #232.

Ping @amontoison


amontoison:
Thanks @gdalle!
Can I finalize the PR for SCT.jl after this one?

gdalle:
Sure, it will be even easier to finalize then, just switch the detector object.
The only thing you have to do is figure out if you want to pass this detector all the way down the call chain, e.g. by making it part of one of your structures.

gdalle:
I suspect your failures are partly due to the ungodly way in which you handle Symbolics. It's not in your Project.toml but it's still a dependency somehow. How does that work?

gdalle:
Oops, for `SymbolicsSparsityDetector` to exist I need the latest version of Symbolics, but they have dropped support for 1.6 a long time ago. If you check the [CI log on main](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/actions/runs/9224320585/job/25379348564), you'll see that you're using an old version of Symbolics.
Note that SparseConnectivityTracer.jl supports Julia 1.6 ;)

amontoison:
I propose to update the compat entry of Julia to v1.9 such we have support for package extensions.
The next release will be breaking in all cases because SCT.jl will be the default.

gdalle:
In that case it probably doesn't make much sense to try hard to support Symbolics again. Maybe you can just rebase your SCT PR on this one, switch the detector and we drop Symbolics once and for all?

amontoison:
> In that case it probably doesn't make much sense to try hard to support Symbolics again. Maybe you can just rebase your SCT PR on this one, switch the detector and we drop Symbolics once and for all?

Yes, let's do that.

gdalle:
Closing in favor of #230

PR #244, opened 2024-06-11
Replace ColPack with SparseMatrixColorings

### LINKED ISSUE #237: Replace ColPack with SparseMatrixColorings?
Hi there @amontoison and friends! It's me again, and I come bearing gifts :sunglasses:
My new package [SparseMatrixColorings.jl](https://github.com/gdalle/SparseMatrixColorings.jl) aims to be a pure-Julia reimplementation of [ColPack](https://github.com/CSCsw/ColPack). It does not yet have every functionality, but it has the essentials. In addition, it offers decompression algorithms, and the test suite is outrageously thorough. Would you be interested in trying it out as a replacement for [ColPack.jl](https://github.com/exanauts/ColPack.jl)?
I haven't yet run benchmarks because there are some things I don't understand about the ColPack.jl interface (see https://github.com/exanauts/ColPack.jl/issues/12 if @michel2323 wants to help), but I expect performance to be rather good.
As an alternative, we could try a full integration with [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) as discussed in #229, but that might be a lot of change at once.

gdalle:
Fix #237 by replacing the dependency on [ColPack.jl](https://github.com/exanauts/ColPack.jl/) with a dependency on [SparseMatrixColorings.jl](https://github.com/gdalle/SparseMatrixColorings.jl).

**Upsides**

Very relevant:

- ColPack is 30k LOCs of unmaintained, unreadable C++ that hasn't been touched in half a decade.
- SparseMatrixColorings.jl is 1.5k LOCs of maintainable, readable pure Julia with good docs and reliable tests.
- SparseMatrixColorings.jl [seems faster](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/237#issuecomment-2160212099) than the upcoming release v0.5 of ColPack.jl, at least for the natural ordering of vertices

A bit less relevant cause it can be fixed with lots of sweat:

- The current release v0.4 of ColPack.jl uses ColPack (C++) in a [_very inefficient way_](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/237#issuecomment-2148116641), with no option for [bipartite coloring](https://github.com/exanauts/ColPack.jl/issues/12)
- We're working with @amontoison to code more efficient bindings in ColPack.jl v0.5, but it is hard to achieve and still rather buggy:
  - https://github.com/exanauts/ColPack.jl/pull/20
  - https://github.com/exanauts/ColPack.jl/issues/21

**Downsides**

- SparseMatrixColorings.jl cannot yet do everything that ColPack (C++) does. See the following issues for tracking:
  - https://github.com/gdalle/SparseMatrixColorings.jl/issues/15
  - https://github.com/gdalle/SparseMatrixColorings.jl/issues/14

**Changes**


src/ad.jl
```diff
@@ -499,36 +499,3 @@ function ADModelNLSBackend(
...
-
-function sparse_matrix_colors(A, alg::ColPackColoration)
-  m, n = size(A)
-  partition_by_rows = alg.partition_choice(m, n)
-  if !isempty(A.nzval)
-    adjA = ColPack.matrix2adjmatrix(A; partition_by_rows = partition_by_rows)
-    CPC = ColPackColoring(adjA, alg.coloring, alg.ordering)
-    colors = get_colors(CPC)
-  else
-    colors = zeros(Int, n)
```
gdalle, line 531: Since I removed this line, the new behavior will be to return a trivial coloring `[1, ..., 1]` if the matrix is empty. Is that breaking?

amontoison: No, it's not breaking. We can allow that.

src/sparse_hessian.jl
```diff
@@ -20,14 +20,15 @@ function SparseADHessian(
...
+  alg::AbstractColoringAlgorithm = GreedyColoringAlgorithm(),
   detector::AbstractSparsityDetector = TracerSparsityDetector(),
   kwargs...,
 ) where {S}
   T = eltype(S)
   H = compute_hessian_sparsity(f, nvar, c!, ncon, detector = detector)
 
-  colors = sparse_matrix_colors(H, alg)
+  # TODO: use ADTypes.symmetric_coloring instead if you have the right decompression
+  colors = ADTypes.column_coloring(H, alg)
```
amontoison, line 31: What should we do to have the right decompression?

amontoison (approved):
Thanks @gdalle!

src/sparse_hessian.jl
gdalle: Use the decompression function which my package includes, just like ColPack itself
https://gdalle.github.io/SparseMatrixColorings.jl/stable/api/#Decompression

src/sparse_hessian.jl
```diff
@@ -20,14 +20,15 @@ function SparseADHessian(
...
-  alg = ColPackColoration(),
+  alg::AbstractColoringAlgorithm = GreedyColoringAlgorithm(),
   detector::AbstractSparsityDetector = TracerSparsityDetector(),
   kwargs...,
 ) where {S}
   T = eltype(S)
   H = compute_hessian_sparsity(f, nvar, c!, ncon, detector = detector)
 
-  colors = sparse_matrix_colors(H, alg)
+  # TODO: use ADTypes.symmetric_coloring instead if you have the right decompression
```
tmigot, line 30: @gdalle Can you make issue on this package that point to this? and the others TODO you added.

tmigot:
@amontoison @gdalle Please squash and not rebase, thanks!

src/sparse_hessian.jl
gdalle: https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/issues/246

Issue #247, opened 2024-06-14
Sparsity detection breaks the norm function on 0.8


DaniGlez:
Hello, I have a couple of optimization problems that do not work with the recent AD upgrades because of (I believe) some interaction of the sparsity detection system with the `norm` function. Here's one MWE that works correctly on v0.7.2 but errors out on v0.8.1 

```julia
using ADNLPModels, LinearAlgebra

F = x -> norm(log.([1 2; 3 4] * x)) - 5
nls = ADNLSModel(F, [1.0, 2.0], 1)
```

I apologize if this is not the correct repository, I'm not familiar enough with the internals of AD to be 100% sure that the issue is in this library and not in a more specific AD library.

gdalle:
TLDR: yes, it's on us (@adrhill and I), but it's easy to fix in a matter of days.

**What is happening?**

Your diagnosis is correct: this bug is linked to sparsity detection. While v0.7 of ADNLPModels.jl relied on Symbolics.jl, v0.8 relies on [SparseConnectivityTracer.jl](https://github.com/adrhill/SparseConnectivityTracer.jl) instead. It is a much simpler (and faster) library, which leverages operator overloading and offers two modes:

- _global_ sparsity detection (the default) tries to output a pattern that doesn't depend on the value of the point `x` at which your Jacobian / Hessian are evaluated
- _local_ sparsity detection adds further optimizations to find more zeros by exploiting the value of `x`, which makes the resulting pattern invalid at another point `x'`

The error tells you that a certain operator called by the `LinearAlgebra.norm` function, namely `ismissing`, is not supported in global mode, and that you would need local mode for it to work. The reason is that global mode forgets the values, and that includes whether or not they are `missing`.

**What can you do about it now?**

Since v0.8, the sparsity detector is not just modified: it is also configurable. Unfortunately that option is not yet documented, despite my annoying requests to @amontoison (see #242). @tmigot if you want to take a stab at it, that would allow @DaniGlez to switch to either one of these detectors:

- `Symbolics.SymbolicSparsityDetector()`: the old world, exactly like in v0.7
- `SparseConnectivityTracer.TracerLocalSparsityDetector()`: a local alternative which will support linear algebra, but beware that the pattern is only valid for the `x` at which you detected it. In most use cases that's not very helpful.

Alternately, you can freeze ADNLPModels.jl to v0.7 for the time being.

**What can we do about it later?**

[...]

PR #249, opened 2024-06-17
[documentation] Explain coloring and detector kwargs in sparse backends

amontoison:
Related to #247 
@gdalle 

I renamed the keyword argument for the coloring algorithm, but I believe it's safe to do so.
Since this option was not documented previously, I don't expect any users to rely on the `alg` option.

### DIFF
--- a/docs/src/mixed.md
+++ b/docs/src/mixed.md
@@ -1,4 +1,4 @@
-# Build an hybrid NLPModel
+# Build a hybrid NLPModel
 
 The package `ADNLPModels.jl` implements the [`NLPModel API`](https://github.com/JuliaSmoothOptimizers/NLPModels.jl) using automatic differentiation (AD) backends.
 It is also possible to build hybrid models that use AD to complete the implementation of a given `NLPModel`.
--- a/docs/src/mixed.md
+++ b/docs/src/mixed.md
@@ -1,4 +1,4 @@
-# Build a hybrid NLPModel
+# Build an hybrid NLPModel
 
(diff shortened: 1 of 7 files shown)


docs/src/mixed.md
```diff
@@ -1,4 +1,4 @@
-# Build a hybrid NLPModel
+# Build an hybrid NLPModel
```
gdalle, line 1: nope, it was correct before

docs/src/sparse.md
```diff
@@ -0,0 +1,46 @@
...
+x = rand(T, 2)
+J = jac(nlp, x)
+```
+
+```@example ex1
+x = rand(T, 2)
+H = hess(nlp, x)
+```
+
+The available backends for sparse derivatives (`SparseADJacobian`, `SparseADHessian` and `SparseReverseADHessian`) have keyword arguments `detector` and `coloring` to specify the sparsity pattern detector and the coloring algorithm, respectively.
```
gdalle, line 34: It's rather weird to call it `SparseADHessian` versus `SparseReverseADHessian`, a better name would be forward-over-forward vs. forward-over-reverse if I understand correctly. But with a switch to DI those distinctions will disappear

gdalle (commented):
Looks good to me, but in general it might be useful to add more `(@ref)` and `(@extref)` (from [DocumenterInterLinks.jl](https://github.com/JuliaLang/Pkg.jl/issues/1585))

docs/src/sparse.md
amontoison: I didn't know that Tangi implement `SparseReverseADHessian`.
The name is not good, we should switch to DI.jl asap!

docs/src/mixed.md
amontoison, line 1:
```suggestion
# Build a hybrid NLPModel
```

Issue #241, opened 2024-06-06
Add performance benchmark


tmigot:
We can get inspiration from this:
https://github.com/lanl-ansi/rosetta-opf/blob/main/nlpmodels.jl / https://discourse.julialang.org/t/ac-optimal-power-flow-in-various-nonlinear-optimization-frameworks/78486

gdalle:
Coming up in https://github.com/gdalle/SparsityDetectionComparison

gdalle:
Here are the comparison results between JuMP, SCT and Symbolics for the first 90% or so of the OptimizationProblems.jl suite. A couple of the last problems take a _really_ long time (probably for Symbolics) and I haven't yet narrowed down which ones. But I think the trend is clear.

![image](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/assets/22795598/47284c82-991d-4ce7-a146-b78b2ae2ec8a)

![image](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/assets/22795598/815ab00b-be5b-41d6-a79a-9428b43870f9)

PR #251, opened 2024-06-20
Add more benchmark: jac and hess

tmigot:
Connected to #241


amontoison (approved):
Maybe we should add a README.md at the root of the folder `benchmark`?

tmigot:
> Maybe we should add a README.md at the root of the folder `benchmark`?

Good idea! I will take care of this.

PR #257, opened 2024-06-24
Optimize sparse AD

amontoison:
- Add a dictionary `dcolors` such that `dcolors[c]` is an array of the components of `vals` that will be computed by the directional derivative associated with color `c`.

We don't need anymore for each color to iterate over all variables to recover which column should be updated.


amontoison:
@tmigot Still not working :(

amontoison:
@tmigot Where are uploaded the results of the benchmarks?

PR #284, opened 2024-08-04
Add an option to provide the sparsity pattern of jacobians and hessians

amontoison:
@jbcaillau

### DIFF
--- a/src/sparse_hessian.jl
+++ b/src/sparse_hessian.jl
@@ -25,8 +25,21 @@ function SparseADHessian(
   detector::AbstractSparsityDetector = TracerSparsityDetector(),
   kwargs...,
 ) where {S}
-  T = eltype(S)
   H = compute_hessian_sparsity(f, nvar, c!, ncon, detector = detector)
+  SparseADHessian(nvar, f, ncon, c!, H; x0, coloring, kwargs...)
+end
+
+function SparseADHessian(
+  nvar,
+  f,
+  ncon,
+  c!,
+  H::SparseMatrixCSC{Bool,Int64};
+  x0::S = rand(nvar),
+  coloring::AbstractColoringAlgorithm = GreedyColoringAlgorithm(),
+  kwargs...,
+) where {S}
+  T = eltype(S)
 
   colors, star_set = symmetric_coloring_detailed(H, coloring)
   ncolors = maximum(colors)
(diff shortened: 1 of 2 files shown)


jbcaillau:
> @jbcaillau

Thanks @amontoison So, provide: 
- sparse AD Jacobian (forward)
- sparse AD Hessian (forward and reverse) ?

amontoison:
> > @jbcaillau
> 
> Thanks @amontoison So, provide:
> 
>     * sparse AD Jacobian (forward)
> 
>     * sparse AD Hessian (forward and reverse) ?

@jbcaillau
The sparsity pattern of Jacobians and Hessians is not related to AD but we must store it for the decompression step.
The AD is just used to recover the value of the nnz coefficients and we can use what we want for the backend.
The supported sparse backend are:
- `SparseADJacobian` -- Forward (Jacobian);
- `SparseADHessian` -- Forward over forward (Hessian);
- `SparseReverseADHessian` -- Forward over Reverse (Hessian).

In the future I would like to only support two backends (`SparseADJacobian` and `SparseADHessian`) with options for the "mode" (forward, reverse, both, forward over reverse, ...) and the "AD package" (Enzyme.jl, ForwardDiff.jl, ReverseDiff.jl, ...).

jbcaillau:
> @jbcaillau The sparsity pattern of Jacobians and Hessians is not related to AD but we must store it for the decompression step. The AD is just used to recover the value of the nnz coefficients and we can use what we want for the backend [...]

@amontoison OK. Thought it was more directly related to this request (example / tuto on how to build an NLPModel)
https://github.com/control-toolbox/CTDirect.jl/issues/183

src/sparse_jacobian.jl
```diff
@@ -22,7 +22,19 @@ function SparseADJacobian(
 ) where {T}
```
tmigot, line 22:
```suggestion
)
```
and adapt line 18

src/sparse_hessian.jl
```diff
@@ -25,8 +25,21 @@ function SparseADHessian(
   detector::AbstractSparsityDetector = TracerSparsityDetector(),
   kwargs...,
 ) where {S}
```
tmigot, line 27:
```suggestion
)
```
and adapt line 23

src/sparse_hessian.jl
```diff
@@ -117,6 +130,19 @@ function SparseReverseADHessian(
   kwargs...,
 ) where {T}
```
tmigot, line 131:
```suggestion
)
```
and adapt line 127v

tmigot (approved):
Thanks @amontoison ! Just a small improvement, I don't think we need the parametric type anymore.

tmigot:
After, this PR I can rehabilitate this PR https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/209 to have such tutorials

amontoison:
@tmigot Do you think that we can add an option when we create an `ADNLPModel` / `ADNLSModel` to provide a sparsity pattern?
An option like `pattern_hessian` / `sp_hessian`.

tmigot:
kwargs given to the ADNLPModels constructor are passed to the ADModelBackend which are passed to each backend.
Would there be a way to modify this PR so that it works?
Maybe add the sparsity pattern as kwargs with default of size (0,0) and then avoid calling the function only if size is different something like this?

I am not against adding kwargs, but ADNLPModels has so many constructors, I would love not adding more code if possible.

amontoison:
Ok, let's not add new keywords and just add an example / tutorial in the documentation to provide the sparsity patterns.

PR #286, opened 2024-08-05
[documentation] Explain how we can provide a sparsity pattern

amontoison:
@tmigot It seems that we can't create an `ADNLPModel` / `ADNLSModel` with an initialized AD backend.
I wanted to create `nlp` with the following constructor:
```julia
nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, acobian_backend=J_backend,hessian_backend=H_backend)
```
but the code [here](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/main/src/ad.jl#L85-L93) wants to call the constructor of an `ADNLPModels.ADBackend`.
We can use `set_adbackend!` as a workaround.

Related to https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/284 and https://github.com/control-toolbox/CTDirect.jl/issues/183
cc @jbcaillau

### DIFF
--- a/docs/src/sparse.md
+++ b/docs/src/sparse.md
@@ -89,11 +89,7 @@ ucon = 0.5 * ones(T, ncon)
 J_backend = ADNLPModels.SparseADJacobian(nvar, f, ncon, c!, J)
 H_backend = ADNLPModels.SparseADHessian(nvar, f, ncon, c!, H)
 
-nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, jacobian_backend=ADNLPModels.EmptyADbackend, hessian_backend=ADNLPModels.EmptyADbackend)
-# nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, matrix_free=true)  # Equivalent to the previous line
-
-set_adbackend!(nlp, jacobian_backend=J_backend)
-set_adbackend!(nlp, hessian_backend=H_backend)
+nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, jacobian_backend=J_backend, hessian_backend=H_backend)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/sparse.md
```diff
@@ -40,6 +40,62 @@ Prior to version 0.8.0, the default detector was `SymbolicSparsityDetector()` fr
 - **Coloring**: A `coloring` must be of type `ADTypes.AbstractColoringAlgorithm`.
 The default algorithm is `GreedyColoringAlgorithm()` from the package `SparseMatrixColorings.jl`.
 
+If the sparsity pattern of the Jacobian of the constraint or the Hessian of the Lagrangian is available, you can directly provide them.
```
amontoison, line 43:
```suggestion
If the sparsity pattern of the Jacobian of the constraints or the Hessian of the Lagrangian is available, you can directly provide them.
```

tmigot:
You could adapt this line https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/c13dbf687d801c0e38b5eebdc4c6ab35127b6113/src/ad.jl#L87 (and the others) to make it work.

Essentially, we expect 3 types as a `jacobian_backend`:
- A `ADBackend` type, and then in the constructor of `ADModelBackend` we call its constructor;
- An `AbstractNLPModel`
- An instantiated `ADBackend` (this is what you want, and is missing right now).

amontoison:
> You could adapt this line
> 
> https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/c13dbf687d801c0e38b5eebdc4c6ab35127b6113/src/ad.jl#L87
> 
> (and the others) to make it work.
> Essentially, we expect 3 types as a `jacobian_backend`:
> 
> * A `ADBackend` type, and then in the constructor of `ADModelBackend` we call its constructor;
> * An `AbstractNLPModel`
> * An instantiated `ADBackend` (this is what you want, and is missing right now).

The issue is what can we do to do the difference between an ADBackend type and an insantiated `ADBackend`?

PR #287, opened 2024-08-06
Allow pre-instantiated backends

tmigot:


tmigot:
@amontoison this should solve https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/286#issuecomment-2269073702 the difference is that an instantiated backend is of type `ADBackend` while in the other case it's just a `DataType`

amontoison (approved):

amontoison:
Thanks @tmigot !!! It works :+1:

PR #289, opened 2024-08-13
Upgrade ADNLPModels.jl for SparseMatrixColoring.jl v0.4.0

amontoison:
cc @gdalle


src/sparse_hessian.jl
```diff
@@ -36,13 +36,17 @@ function SparseADHessian(
...
   T = eltype(S)
 
-  colors, star_set = symmetric_coloring_detailed(H, coloring)
-  ncolors = maximum(colors)
+  problem = ColoringProblem(; structure=:symmetric, partition=:column)
+  result = coloring(H, problem, coloring, decompression_eltype=T)
+  colors = column_colors(result)
+  groups = column_groups(result)
+  ncolors = length(groups)
+  star_set = result.star_set
```
gdalle, line 49: the `star_set` is no longer part of the public API, you have to rely on `decompress` instead

src/sparse_hessian.jl
```diff
@@ -140,11 +144,15 @@ function SparseReverseADHessian(
...
   kwargs...,
 ) where {T}
-  colors, star_set = symmetric_coloring_detailed(H, coloring)
-  ncolors = maximum(colors)
+  problem = ColoringProblem(; structure=:symmetric, partition=:column)
+  result = coloring(H, problem, coloring, decompression_eltype=T)
+  colors = column_colors(result)
+  groups = column_groups(result)
+  ncolors = length(groups)
+  star_set = result.star_set
```
gdalle, line 155: same remark here about the `star_set`

gdalle (changes requested):

src/sparse_hessian.jl
amontoison: I don't want to always store all the compressed columns if I use direct recovery.
It could be prohibitive in terms of storage.
I just want to use `decompress` for the acyclic coloring and use star coloring as a backup if memory becomes an issue.
Also `decompress` don't support an option to update just one triangle right now, which is important for me.

docs/src/sparse.md
```diff
@@ -31,14 +31,14 @@ x = rand(T, 2)
...
+The available backends for sparse derivatives (`SparseADJacobian`, `SparseADHessian` and `SparseReverseADHessian`) have keyword arguments `detector` and `coloring_options` to specify the sparsity pattern detector and the coloring options, respectively.
 
 - **Detector**: A `detector` must be of type `ADTypes.AbstractSparsityDetector`.
 The default detector is `TracerSparsityDetector()` from the package `SparseConnectivityTracer.jl`.
 Prior to version 0.8.0, the default detector was `SymbolicSparsityDetector()` from `Symbolics.jl`.
 
-- **Coloring**: A `coloring` must be of type `ADTypes.AbstractColoringAlgorithm`.
-The default algorithm is `GreedyColoringAlgorithm()` from the package `SparseMatrixColorings.jl`.
+- **Coloring**: A `coloring_options` must be of type `ADTypes.AbstractColoringAlgorithm`.
+The default algorithm is `GreedyColoringAlgorithm{:direct}()` from the package `SparseMatrixColorings.jl`.
```
gdalle, line 41: Not worth specifying that it must subtype `ADTypes.AbstractColoringAlgorithm`, because the one from SparseMatrixColorings is the only subtype we support. Better to say that it must be an instance of `SparseMatrixColorings.GreedyColoringAlgorithm`

docs/src/sparse.md
```diff
@@ -31,14 +31,14 @@ x = rand(T, 2)
 H = hess(nlp, x)
 ```
 
-The available backends for sparse derivatives (`SparseADJacobian`, `SparseADHessian` and `SparseReverseADHessian`) have keyword arguments `detector` and `coloring` to specify the sparsity pattern detector and the coloring algorithm, respectively.
+The available backends for sparse derivatives (`SparseADJacobian`, `SparseADHessian` and `SparseReverseADHessian`) have keyword arguments `detector` and `coloring_options` to specify the sparsity pattern detector and the coloring options, respectively.
```
gdalle, line 34: Why not name it `coloring_algorithm`?

src/sparse_hessian.jl
```diff
@@ -237,76 +227,77 @@ function NLPModels.hess_structure_residual!(
...
-      b.res .= view(b.Hvp, (ncon + 1):(ncon + nvar))
+  # SparseMatrixColorings.jl requires a SparseMatrixCSC for the decompression
+  A = SparseMatrixCSC(b.nvar, b.nvar, b.colptr, b.rowval, b.nzval)
 
-      # Store in `vals` the nonzeros of each column of the Hessian computed with color `icol`
-      for (row, k) in dcolor_icol
-        vals[k] = b.res[row]
-      end
+  groups = column_groups(b.result_coloring)
+  for (icol, cols) in enumerate(groups)
```
gdalle, line 249: Why is `A` not used with `decompress!` in this loop?

gdalle (changes requested):

docs/src/sparse.md
gdalle: @amontoison this is important

amontoison: Done

PR #299, opened 2024-08-17
Remove ct-registry for OptimalControl.jl

ocots:
I have remove in `.github/workflows/Breakage.yml` the following lines:

```yml
if [ $URL == "control-toolbox/OptimalControl.jl" ]; then
    julia -e 'using Pkg; Pkg.Registry.add(RegistrySpec(url = "https://github.com/control-toolbox/ct-registry.git"))'
fi
```

since now OptimalControl.jl is in the General registry. There no need to add `ct-registry` anymore.

### DIFF
--- a/.github/workflows/Breakage.yml
+++ b/.github/workflows/Breakage.yml
@@ -56,9 +56,6 @@ jobs:
           set -v
           mkdir -p ./pr
           echo "${{ github.event.number }}" > ./pr/NR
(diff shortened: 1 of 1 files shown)


ocots:
I don't know if with the PR, we will have these checks: https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/pull/294#issuecomment-2294324336

amontoison:
Thanks @ocots!

PR #302, opened 2024-09-11
 Add new backends with DifferentiationInterface.jl

amontoison:
Add a  new backend based on `DifferentiationInterface.jl`.


docs/src/backend.md
```diff
@@ -10,10 +10,10 @@ The functions used internally to define the NLPModel API and the possible backen
 | Functions | FowardDiff backends | ReverseDiff backends | Zygote backends | Enzyme backend | Sparse backend |
 | --------- | ------------------- | -------------------- | --------------- | -------------- | -------------- |
 | `gradient` and `gradient!` | `ForwardDiffADGradient`/`GenericForwardDiffADGradient` | `ReverseDiffADGradient`/`GenericReverseDiffADGradient` | `ZygoteADGradient` | `EnzymeADGradient` | -- |
-| `jacobian` | `ForwardDiffADJacobian` | `ReverseDiffADJacobian` | `ZygoteADJacobian` | -- | `SparseADJacobian` |
-| `hessian` | `ForwardDiffADHessian` | `ReverseDiffADHessian` | `ZygoteADHessian` | -- | `SparseADHessian`/`SparseReverseADHessian` |
```
amontoison, line 14: @tmigot Do we have two functions `hessian`, one of the objective and one for the Lagrangian?
I was not sure so I commented to code.
The version for the Lagrangian can't be easily implemented.

Project.toml
```diff
@@ -15,6 +16,7 @@ SparseMatrixColorings = "0a514795-09f3-496d-8182-132a7b665d35"
 
 [compat]
 ADTypes = "1.2.1"
+DifferentiationInterface = "0.5.17"
 ForwardDiff = "0.9.0, 0.10.0"
 NLPModels = "0.18, 0.19, 0.20, 0.21"
 Requires = "1"
```
tmigot, line 22:
```suggestion
```
we used Require for Zygote and Enzyme

docs/src/backend.md
tmigot: It's only one, so basically with this change we would no longer be able to use Hessian for Enzyme and Zygote.

src/di.jl
```diff
@@ -0,0 +1,195 @@
...
+      kwargs...,
+    )
+      backend = $fbackend()
+      extras = DifferentiationInterface.prepare_gradient(f, backend, x0)
+      return $ADGradient(backend, extras)
+    end
+
+    function gradient(b::$ADGradient, f, x)
+      g = DifferentiationInterface.gradient(f, b.backend, x, b.extras)
+      # g = DifferentiationInterface.gradient(f, b.extras, b.backend, x)
```
tmigot, line 25:
```suggestion
      # g = DifferentiationInterface.gradient(f, b.extras, b.backend, x)
```
?? Please remove dead code

src/di.jl
```diff
@@ -0,0 +1,195 @@
...
+    function jacobian(b::$ADJacobian, f, x)
+      J = DifferentiationInterface.jacobian(f, b.backend, x, b.extras)
+      # J = DifferentiationInterface.jacobian(f, b.extras, b.backend, x)
+      return J
+    end
+
+  end
+end
+
+# for (ADHessian, fbackend) in ((:EnzymeADHessian, :AutoEnzyme),
```
tmigot, line 134: We should avoid having commented code like this in main. You can always open draft PRs for this.

test/runtests.jl
```diff
@@ -3,6 +3,7 @@ using SparseMatrixColorings
 using ADNLPModels, ManualNLPModels, NLPModels, NLPModelsModifiers, NLPModelsTest
 using ADNLPModels:
   gradient, gradient!, jacobian, hessian, Jprod!, Jtprod!, directional_second_derivative, Hvprod!
+import Enzyme, Zygote
```
tmigot, line 6:
```suggestion
using Enzyme, Zygote
```

test/runtests.jl
```diff
@@ -96,52 +111,49 @@ function test_getter_setter(nlp)
...
+#   ADNLPModels.ZygoteADJtprod(),
+#   ADNLPModels.ZygoteADJacobian(0),
+#   ADNLPModels.ZygoteADHessian(0),
+#   ADNLPModels.ForwardDiffADGHjvprod(),
+#   ADNLPModels.EmptyADbackend(),
+#   ADNLPModels.EmptyADbackend(),
+#   ADNLPModels.EmptyADbackend(),
+#   ADNLPModels.EmptyADbackend(),
+#   ADNLPModels.EmptyADbackend(),
+# )
```
tmigot, line 127:
```suggestion
```

test/runtests.jl
```diff
@@ -96,52 +111,49 @@ function test_getter_setter(nlp)
...
+#     @test_throws ArgumentError Jtprod!(
+#       adbackend.jtprod_backend,
+#       [1.0],
+#       [1.0],
+#       identity,
+#       [1.0],
+#       Val(:c),
+#     )
+#   end
+# end
```
tmigot, line 153:
```suggestion
```

test/runtests.jl
```diff
@@ -96,52 +111,49 @@ function test_getter_setter(nlp)
...
 
 # Test the argument error without loading the packages
```
tmigot, line 155:
```suggestion
```

test/runtests.jl
```diff
@@ -96,52 +111,49 @@ function test_getter_setter(nlp)
...
-test_autodiff_backend_error()
-
-# Automatically loads the code for Zygote with Requires
-import Zygote
+# test_autodiff_backend_error()
```
tmigot, line 156:
```suggestion
```

test/script_OP.jl
```diff
@@ -55,79 +52,3 @@ for pb in names
     continue
   end
 end
-
-#=
-ADNLPModels.EmptyADbackend(args...; kwargs...) = ADNLPModels.EmptyADbackend()
```
tmigot, line 60: Does this PR fix this?

tmigot (changes requested):
Thanks @amontoison for the PR. I have mixed feelings with it. On one side it is progress, on the other side we are loosing Hessian backend for Enzyme and Zygote.

How far are we from making it fully compatible?

docs/src/backend.md
amontoison: Well, it was broken in the case. The user provided `f` and `c` but only the Hessian of `f` was returned.

src/di.jl
amontoison: I forgot to add a comment, it's not dead code but a the code for the release 0.6.0 of DI.jl.
I don't would like to use it with the next major version of ADNLPModels.jl.

src/di.jl
amontoison: @tmigot 
I commented it because I waited for an answer of you on the function `hessian`.
Your `hessian` of Zygote seems wrong:
https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/main/src/zygote.jl#L67-L84

test/runtests.jl
amontoison: Not a good idea, each AD package could export their functions `gradient`, `jacobian`, `hessian`, etc...
They will collapse with a `using`.

test/script_OP.jl
amontoison: Fix what Tangi?
Can you elaborate?

amontoison:
> It's only one, so basically with this change we would no longer be able to use Hessian for Enzyme and Zygote.

We can but only for unconstrained problems.
I wanted to remove what was not working before.
The user will no longer be able to use an incorrect Hessian, which is better.

docs/src/backend.md
tmigot: Oh, you are right. Ok.

src/di.jl
tmigot: Ok, but still let's not have commented lines like this. You can add this in an issue or a draft PR.

test/script_OP.jl
tmigot: This comment block contains a list of problems that failed

amontoison:
@gdalle May I ask you to check what I did wrong in the file `di.jl`?
I have different errors with `buildkite`: https://buildkite.com/julialang/adnlpmodels-dot-jl/builds/243

gdalle:
It looks like the problem comes from forgetting to import the function `grad`?

gdalle:
@dpo can you give me acces to the repo so that I may help with this and future PRs?

docs/src/backend.md
```diff
@@ -1,23 +1,25 @@
 # How to switch backend in ADNLPModels
 
 `ADNLPModels` allows the use of different backends to compute the derivatives required within NLPModel API.
-It uses `ForwardDiff.jl`, `ReverseDiff.jl`, and more via optional depencies.
+It uses `ForwardDiff.jl`, `ReverseDiff.jl`, and more via optional dependencies.
 
 The backend information is in a structure [`ADNLPModels.ADModelBackend`](@ref) in the attribute `adbackend` of a `ADNLPModel`, it can also be accessed with [`get_adbackend`](@ref).
 
 The functions used internally to define the NLPModel API and the possible backends are defined in the following table:
```
gdalle, line 8: Why not just switch fully to the ADTypes specification? You're gonna run into trouble translating symbols into `AbstractADType` objects

gdalle: And the symbols don't allow you to set parameters like
- the number of chunks in ForwardDiff
- the tape compilation in ReverseDiff
- aspects of the mode in Enzyme

docs/src/backend.md
```diff
@@ -1,23 +1,25 @@
...
+| $\nabla^2 \mathcal{L}(x)$   | `ForwardDiffADHessian`    | `ReverseDiffADHessian`      | `EnzymeADHessian`     | `ZygoteADHessian`     | `MooncakeADHessian`     | `DiffractorADHessian`     | `TrackerADHessian`     | `SymbolicsADHessian`     | `ChainRulesADHessian`            | `FastDifferentiationADHessian`    | `FiniteDiffADHessian`      | `FiniteDifferencesADHessian`    | `PolyesterForwardDiffADHessian`   |
+
+$\mathcal{L}(x)$ denotes the Lagrangian $f(x) + \lambda^T c(x)$.
+Except for the backends based on `ForwardDiff.jl` and `ReverseDiff.jl`, all other backends require the associated AD package to be manually installed by the user to work.
+Note that the Jacobians and Hessians computed by the backends above are dense.
+The backends `SparseADJacobian`, `SparseADHessian`, and `SparseReverseADHessian` should be used instead if sparse Jacobians and Hessians are required.
```
gdalle, line 22: Same remark for sparse AD, using `AutoSparse` seems more flexible?

src/di.jl
```diff
@@ -0,0 +1,256 @@
+for (ADGradient, fbackend) in ((:EnzymeADGradient              , :AutoEnzyme              ),
+                               (:ZygoteADGradient              , :AutoZygote              ),
+                             # (:ForwardDiffADGradient         , :AutoForwardDiff         ),
+                             # (:ReverseDiffADGradient         , :AutoReverseDiff         ),
+                               (:MooncakeADGradient            , :AutoMooncake            ),
```
gdalle, line 5: The `AutoMooncake` constructor requires a keyword, like so:
```julia
AutoMooncake(; config=nothing)
```

src/di.jl
```diff
@@ -0,0 +1,256 @@
...
+                               (:DiffractorADGradient          , :AutoDiffractor          ),
+                               (:TrackerADGradient             , :AutoTracker             ),
+                               (:SymbolicsADGradient           , :AutoSymbolics           ),
+                               (:ChainRulesADGradient          , :AutoChainRules          ),
```
gdalle, line 9: The `AutoChainRules` constructor requires a keyword, like so:
```julia
AutoChainRules(; ruleconfig=Zygote.ZygoteRuleConfig())
```

src/di.jl
```diff
@@ -0,0 +1,256 @@
...
+                               (:FastDifferentiationADGradient , :AutoFastDifferentiation ),
+                               (:FiniteDiffADGradient          , :AutoFiniteDiff          ),
+                               (:FiniteDifferencesADGradient   , :AutoFiniteDifferences   ),
```
gdalle, line 12: The `AutoFiniteDifferences` constructor requires a keyword, like so:
```julia
AutoFiniteDifferences(; fdm=FiniteDifferences.central_fdm(3, 1))
```

src/di.jl
```diff
@@ -0,0 +1,256 @@
...
+
+    function $ADGradient(
+      nvar::Integer,
+      f,
+      ncon::Integer = 0,
+      c::Function = (args...) -> [];
+      x0::AbstractVector = rand(nvar),
+      kwargs...,
+    )
+      backend = $fbackend()
```
gdalle, line 29: This will fail for the three backends mentioned above. And for all other backends, this prevents you from setting any options, which was the goal of ADTypes.jl to begin with: see https://github.com/SciML/ADTypes.jl?tab=readme-ov-file#why-should-ad-users-adopt-this-standard

src/di.jl
```diff
@@ -0,0 +1,256 @@
...
+                           # (:ReverseDiffADHvprod         , :AutoReverseDiff         ),
+                             (:MooncakeADHvprod            , :AutoMooncake            ),
+                             (:DiffractorADHvprod          , :AutoDiffractor          ),
+                             (:TrackerADHvprod             , :AutoTracker             ),
+                             (:SymbolicsADHvprod           , :AutoSymbolics           ),
+                             (:ChainRulesADHvprod          , :AutoChainRules          ),
+                             (:FastDifferentiationADHvprod , :AutoFastDifferentiation ),
+                             (:FiniteDiffADHvprod          , :AutoFiniteDiff          ),
+                             (:FiniteDifferencesADHvprod   , :AutoFiniteDifferences   ),
+                             (:PolyesterForwardDiffADHvprod, :AutoPolyesterForwardDiff))
```
gdalle, line 187: Diffractor, Mooncake, Tracker and ChainRules probably don't work in second order.
FiniteDiff and FiniteDifferences might give you inaccurate results depending on their configuration (https://github.com/gdalle/DifferentiationInterface.jl/issues/78)

src/di.jl
```diff
@@ -0,0 +1,256 @@
...
+
+    function Hvprod!(b::$ADHvprod, Hv, f, x, v, ::Val)
+      DifferentiationInterface.hvp!(f, Hv, b.prep, b.backend, x, v)
+      return Hv
+    end
+
+  end
+end
+
+for (ADHessian, fbackend) in ((:EnzymeADHessian              , :AutoEnzyme              ),
```
gdalle, line 217: Same remark as for HVP about backends incompatible with second order

src/predefined_backend.jl
```diff
@@ -43,7 +43,217 @@ generic = Dict(
...
+  :hprod_backend => ForwardDiffADHvprod,
+  :jacobian_backend => ForwardDiffADJacobian,
+  :hessian_backend => ForwardDiffADHessian,
+  :ghjvprod_backend => EmptyADbackend,
+  :jprod_residual_backend => ForwardDiffADJprod,
+  :jtprod_residual_backend => ForwardDiffADJtprod,
+  :hprod_residual_backend => ForwardDiffADHvprod,
+  :jacobian_residual_backend => ForwardDiffADJacobian,
+  :hessian_residual_backend => ForwardDiffADHessian
+)
```
gdalle, line 59: The goal of DifferentiationInterface is to save a lot people a lot code. However, this PR ends up adding more lines than it removes, precisely because of this kind of disjunction.
@Vaibhavdixit2 how did you handle the choice of backend for each operator in OptimizationBase.jl?

test/backend/utils.jl
```diff
@@ -0,0 +1,12 @@
+using OptimizationProblems
+
```
gdalle, line 2:
```suggestion
using NLPModels
```

gdalle (changes requested):
My gut feeling is that if you want to switch to DI, you have to switch to ADTypes as well and stop doing this cumbersome translation between symbol and backend objects. What do you think?

dpo:
@gdalle I invited you. Thank you for your work here!!!

gdalle:
@amontoison what do you think about moving away from symbols here?

amontoison:
> @amontoison what do you think about moving away from symbols here?

It depends on the alternatives, Right now, it's useful to specify that we want optimized backends with `:optimized` or only matrix-free backends `:matrix_free`.
But if `Enzyme.jl` is stable enough, we could drop `:optimized` and use a boolean for matrix-free backends.

gdalle:
If I'm not mistaken there are two levels here:
- the interface you present to the user (`:optimized`, `:matrix_free`)
- the way you represent the backends internally

Right now you base all of the internal representations on `Symbol`s. But as explained [here](https://github.com/SciML/ADTypes.jl?tab=readme-ov-file#why-should-ad-users-adopt-this-standard), the whole reason for ADTypes was to move beyond `Symbol`s towards full-fledged types that are 1) more expressive and 2) dispatchable. That's why I was suggesting a similar move here. It doesn't stop you from offering `:optimized` autodiff options in the front end if you like

amontoison:
Do you have an example of what you suggest?

gdalle:
I could try to show you in an alternative PR

gdalle:
Okay it is a bit hard to submit a PR since there would be a lot of things to rewrite and I don't understand what each part does. But essentially I was imagining something like this:

```julia
using ADTypes
using DifferentiationInterface
using LinearAlgebra
using SparseMatrixColorings
using SparseConnectivityTracer
import ForwardDiff, ReverseDiff

function DefaultAutoSparse(backend::AbstractADType)
    return AutoSparse(
        backend;
        sparsity_detector=TracerSparsityDetector(),
        coloring_algorithm=GreedyColoringAlgorithm(),
    )
end

struct ADModelBackend
    gradient_backend
    hprod_backend
    jprod_backend
    jtprod_backend
    jacobian_backend
    hessian_backend
end

struct ADModelBackendPrep
    gradient_prep
    hprod_prep
    jprod_prep
    jtprod_prep
    jacobian_prep
    hessian_prep
end

function ADModelBackend(forward_backend::AbstractADType, reverse_backend::AbstractADType)
    @assert ADTypes.mode(forward_backend) isa
        Union{ADTypes.ForwardMode,ADTypes.ForwardOrReverseMode}
    @assert ADTypes.mode(reverse_backend) isa
        Union{ADTypes.ReverseMode,ADTypes.ForwardOrReverseMode}

    gradient_backend = reverse_backend
    hprod_backend = SecondOrder(forward_backend, reverse_backend)
    jprod_backend = forward_backend
    jtprod_backend = reverse_backend
    jacobian_backend = DefaultAutoSparse(forward_backend)  # or a size-dependent heuristic
    hessian_backend = DefaultAutoSparse(SecondOrder(forward_backend, reverse_backend))

    return ADModelBackend(
        gradient_backend,
        hprod_backend,
        jprod_backend,
        jtprod_backend,
        jacobian_backend,
        hessian_backend,
    )
end

function ADModelBackendPrep(
    admodel_backend::ADModelBackend,
    obj::Function,
    cons::Function,
    lag::Function,
    x::AbstractVector,
)
    (;
        gradient_backend,
        hprod_backend,
        jprod_backend,
        jtprod_backend,
        jacobian_backend,
        hessian_backend,
    ) = admodel_backend

    c = cons(x)
    λ = similar(c)

    dx = similar(x)
    dc = similar(c)

    gradient_prep = prepare_gradient(obj, gradient_backend, x)
    hprod_prep = prepare_hvp(lag, hprod_backend, x, (dx,), Constant(λ))
    jprod_prep = prepare_pushforward(cons, jprod_backend, x, (dx,))
    jtprod_prep = prepare_pullback(cons, jtprod_backend, x, (dc,))
    jacobian_prep = prepare_jacobian(cons, jacobian_backend, x)
    hessian_prep = prepare_hessian(lag, hessian_backend, x, Constant(λ))

    return ADModelBackendPrep(
        gradient_prep, hprod_prep, jprod_prep, jtprod_prep, jacobian_prep, hessian_prep
    )
end

admodel_backend = ADModelBackend(AutoForwardDiff(), AutoReverseDiff())

obj(x) = sum(x)
cons(x) = abs.(x)
lag(x, λ) = obj(x) + dot(λ, cons(x))

admodel_backend_prep = ADModelBackendPrep(admodel_backend, obj, cons, lag, rand(3));
```

PR #327, opened 2025-02-01
Provide an option to exclude some AD backends

### LINKED ISSUE #324: Add an option to provide the list of backends that we want
If we just need `grad`, `obj`, `jacobian`, `hessian`, we would to not create backends for the other ones (`jprod`, `jtprod`, `hprod`, `ghjvprod`, etc...)

@tmigot Do you have an idea on what is the best way to do that?

amontoison:
close #324 
cc @jbcaillau @PierreMartinon
It will be easier for you to exclude some `ADbackend` with the option `exclude_backend`.
```julia
using ADNLPModels, NLPModels

f(x) = (x[1] - 1)^2
T = Float64
x0 = T[-1.2; 1.0]
lvar, uvar = zeros(T, 2), ones(T, 2)
lcon, ucon = -T[0.5], T[0.5]
c!(cx, x) = begin
  cx[1] = x[2]
  return cx
end
nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, backend = :optimized,
                  excluded_backend=[:jprod_backend, :jtprod_backend, :hprod_backend, :ghjvprod_backend])
```
```julia
ADNLPModel - Model with automatic differentiation backend ADModelBackend{
  ReverseDiffADGradient,
  EmptyADbackend,
  EmptyADbackend,
  EmptyADbackend,
  SparseADJacobian,
  SparseReverseADHessian,
  EmptyADbackend,
}
```

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -16,6 +16,7 @@ The default constructors are
     ADModelNLSBackend(nvar, F!, nequ, ncon = 0, c = (args...) -> []; show_time::Bool = false, kwargs...)
 
 If `show_time` is set to `true`, it prints the time used to generate each backend.
+If `excluded_backend` is not an empty array `Symbol[]`, the excluded backends will be set to `EmptyADbackend`.
 
 The remaining `kwargs` are either the different backends as listed below or arguments passed to the backend's constructors:
   - `gradient_backend = ForwardDiffADGradient`;
@@ -54,9 +55,10 @@ function ADModelBackend(
   backend::Symbol = :default,
   matrix_free::Bool = false,
   show_time::Bool = false,
-  gradient_backend = get_default_backend(:gradient_backend, backend),
-  hprod_backend = get_default_backend(:hprod_backend, backend),
-  hessian_backend = get_default_backend(:hessian_backend, backend, matrix_free),
+  excluded_backend::Vector{Symbol} = Symbol[],
+  gradient_backend = get_default_backend(:gradient_backend, backend; excluded_backend),
+  hprod_backend = get_default_backend(:hprod_backend, backend; excluded_backend),
+  hessian_backend = get_default_backend(:hessian_backend, backend, matrix_free; excluded_backend),
   kwargs...,
 )
   c! = (args...) -> []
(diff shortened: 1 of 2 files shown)


tmigot:
The current way to achieve this is to set the mentioned backend to `EmptyADbackend`, but if you guys think that might be useful why not.

src/ad.jl
```diff
@@ -16,6 +16,7 @@ The default constructors are
     ADModelNLSBackend(nvar, F!, nequ, ncon = 0, c = (args...) -> []; show_time::Bool = false, kwargs...)
 
 If `show_time` is set to `true`, it prints the time used to generate each backend.
+If `excluded_backend` is not an empty array `Symbol[]`, the excluded backends will be set to `EmptyADbackend`.
```
tmigot, line 19:
```suggestion
`excluded_backend` is the list of backends that are not used and will be set to `EmptyADbackend`.
```

tmigot (changes requested):
Looks overall good, thanks @amontoison !
Could you also update the section https://jso.dev/ADNLPModels.jl/dev/performance/ for the doc and maybe add a very short unit test?

amontoison:
> The current way to achieve this is to set the mentioned backend to `EmptyADbackend`, but if you guys think that might be useful why not.

I forgot that we can do that!
I found it in the documentation: https://jso.dev/ADNLPModels.jl/dev/performance/#Use-only-the-needed-operations

I don't think this PR will bring any advantage. I will just update the documentation to specify the backends needed for Ipopt / Knitro / MadNLP.
```julia
using ADNLPModels, NLPModels

f(x) = (x[1] - 1)^2
T = Float64
x0 = T[-1.2; 1.0]
lvar, uvar = zeros(T, 2), ones(T, 2)
lcon, ucon = -T[0.5], T[0.5]
c!(cx, x) = begin
  cx[1] = x[2]
  return cx
end
nlp = ADNLPModel!(f, x0, lvar, uvar, c!, lcon, ucon, backend = :optimized,
                         jprod_backend = ADNLPModels.EmptyADbackend,
                         jtprod_backend = ADNLPModels.EmptyADbackend,
                         hprod_backend = ADNLPModels.EmptyADbackend(),
                         ghjvprod_backend = ADNLPModels.EmptyADbackend())
```

amontoison:
I updated the documentation in #328.

PR #328, opened 2025-02-01
[documentation] Update the performance tips

amontoison:
@jbcaillau

### DIFF
--- a/docs/src/performance.md
+++ b/docs/src/performance.md
@@ -125,8 +125,8 @@ This means that we can set all other backends to `ADNLPModels.EmptyADbackend`.
 
 ```@example ex2
 nlp = ADNLPModel!(f, x0, c_in, lcon, ucon, jprod_backend = ADNLPModels.EmptyADbackend,
-                  jtprod_backend = ADNLPModels.EmptyADbackend, hprod_backend = ADNLPModels.EmptyADbackend(),
-                  ghjvprod_backend = ADNLPModels.EmptyADbackend(), show_time = true)
+                  jtprod_backend = ADNLPModels.EmptyADbackend, hprod_backend = ADNLPModels.EmptyADbackend,
+                  ghjvprod_backend = ADNLPModels.EmptyADbackend, show_time = true)
 ```
 
 ## Benchmarks
--- a/docs/src/performance.md
+++ b/docs/src/performance.md
@@ -1,6 +1,6 @@
 # Performance tips
 
-The package `ADNLPModels.jl` is designed to easily model optimization problems andto allow an efficient access to the [`NLPModel API`](https://github.com/JuliaSmoothOptimizers/NLPModels.jl).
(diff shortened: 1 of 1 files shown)


docs/src/performance.md
jbcaillau, line 1: @amontoison nice. regarding the syntax, it's getting a bit heavy. though it does not seem to be standard in Julia yet, I'd advocate sth like what used to be done, *e.g.*, in matlab: build iteratively the options using an *ad hoc* constructor, then, pass it to the code. or having a dedicated keyword for such "empty" backends...

amontoison: @tmigot Any idea of how we can design that with the current API?

docs/src/performance.md
```diff
@@ -120,6 +120,15 @@ or, equivalently, using the `matrix_free` keyword argument
 nlp = ADNLPModel!(f, x0, c_in, lcon, ucon, show_time = true, matrix_free = true)
 ```
 
+More classic nonlinear optimization solvers like [Ipopt.jl](https://github.com/jump-dev/Ipopt.jl), [KNITRO.jl](https://github.com/jump-dev/KNITRO.jl), or [MadNLP.jl](https://github.com/MadNLP/MadNLP.jl) only require the gradient and sparse Jacobians and Hessians.
+This means that we can set all other backends to `ADNLPModels.EmptyADbackend`.
+
+```@example ex2
+nlp = ADNLPModel!(f, x0, c_in, lcon, ucon, jprod_backend = ADNLPModels.EmptyADbackend,
+                  jtprod_backend = ADNLPModels.EmptyADbackend, hprod_backend = ADNLPModels.EmptyADbackend(),
```
amontoison, line 128:
```suggestion
                  jtprod_backend = ADNLPModels.EmptyADbackend, hprod_backend = ADNLPModels.EmptyADbackend,
```

docs/src/performance.md
```diff
@@ -120,6 +120,15 @@ or, equivalently, using the `matrix_free` keyword argument
...
+                  ghjvprod_backend = ADNLPModels.EmptyADbackend(), show_time = true)
```
amontoison, line 129:
```suggestion
                  ghjvprod_backend = ADNLPModels.EmptyADbackend, show_time = true)
```

tmigot (approved):
Thanks @amontoison , LGTM

docs/src/performance.md
tmigot: @jbcaillau do you mean having something like:
```
ADNLPModel(f, x0, c, lcon, ucon, gradient_backend)
ADNLPModel(f, x0, c, lcon, ucon, gradient_backend, jacobian_backend)
ADNLPModel(f, x0, c, lcon, ucon, gradient_backend, jacobian_backend, jprod_backend, jtprod_backend)
ADNLPModel(f, x0, c, lcon, ucon, gradient_backend, jacobian_backend, jprod_backend, jtprod_backend, hessian_backend)
...
```
I find it kind of heavy and it will not be 100% clear or do you mean something different ?

Note that in the code right now, all these backends objects all are subtypes of ADBackend, so we don't have ways to distinguish between different types of backends without keywords.

jbcaillau: it s ok like that

Issue #347, opened 2025-05-05
Type instability in ADNLPModels


MaxenceGollier:
Hello, 
I am trying to implement an efficient optimization solver. In order to test performance I use ADNLPModels from `OptimizationProblems.jl`, I am currently trying to catch type instabilities/allocations using `JET.jl` and I have found that a lot of issues where coming from this package. 
This is weird because even for very simple operations Jet throws type instabilities at me, I guess I am doing something wrong but I can not find my error. I am not sure whether this belongs here or if I should post it on `OptimizationProblems.jl`. 
Anyway, here is a minimal working example: 
```
using JET, ADNLPModels, NLPModels, OptimizationProblems, OptimizationProblems.ADNLPProblems
nlp = bt1()
x = zeros(2)
@report_opt obj(nlp, x)
```
which results in 
```
═════ 1 possible error found ═════
┌ obj(nlp::ADNLPModel{Float64, Vector{Float64}, Vector{Int64}}, x::Vector{Float64}) @ ADNLPModels C:\Users\mgoll\.julia\packages\ADNLPModels\ELm6c\src\nlp.jl:538
│ runtime dispatch detected: %22::Any(x::Vector{Float64})::Any
```
Additionnally, this causes allocations as
```
 @allocated obj(nlp, x)
 @allocated obj(nlp, x)
```
results in `16` which is quite small but I might have to call this function and other ADNLPModels related functions very often.

Let me know what you think, thank you !

tmigot:
Mmmhhh, it is possible. Can you try adding a parametric type for the function `f` here:
```
mutable struct ADNLPModel{T, S, Si} <: AbstractADNLPModel{T, S}
  meta::NLPModelMeta{T, S}
  counters::Counters
  adbackend::ADModelBackend

  # Functions
  f

  clinrows::Si
  clincols::Si
  clinvals::S

  c!
end
```
?

I sometimes test like this:
```
nlp = OptimizationProblems.ADNLPProblems.bt1()
function check_obj_allocs(nlp)
    x = zeros(2)
    obj(nlp, x)
    return nothing
end
@allocated check_obj_allocs(nlp)
@allocated check_obj_allocs(nlp)
```
if that works we should add an allocation test. I thought it existed for residuals and constraints functions but can't find it in the tests, so maybe I was doing it opportunistically.

MaxenceGollier:
Though it did work for the objective functions there are still lots of type instabilities here and there. 
[...]

PR #352, opened 2025-06-05
Fix type instabilities in ADNLPModels.jl

MaxenceGollier:
#347 

@amontoison @tmigot This is where I am so far. 

Attached are my benchmark results...
[grad_benchmarks.zip](https://github.com/user-attachments/files/20606813/grad_benchmarks.zip)

### DIFF
--- a/src/ADNLPModels.jl
+++ b/src/ADNLPModels.jl
@@ -160,7 +160,7 @@ function ADNLSModel!(model::AbstractNLSModel; kwargs...)
   end
 end
 
-export get_adbackend
+export get_adbackend, set_adbackend
 
 """
     get_c(nlp)
@@ -248,7 +248,7 @@ get_adbackend(nlp::ADModel) = nlp.adbackend
     new_nlp = set_adbackend(nlp; kwargs...)
 
 Create a copy of nlp that replaces the current `adbackend` with `new_adbackend` or instantiate a new one with `kwargs`, see `ADModelBackend`.
-By default, the setter with kwargs will reuse existing backends.
+By default, the setter with keyword arguments will reuse existing backends.
 """
 function _set_adbackend(nlp::ADM, new_adbackend::ADModelBackend) where{ADM}
   values = [f == :adbackend ? new_adbackend : getfield(nlp, f) for f in fieldnames(ADM)]
--- a/docs/src/backend.md
+++ b/docs/src/backend.md
@@ -96,26 +96,27 @@ nlp = ADNLPModel(f, x0)
 get_adbackend(nlp) # returns the `ADModelBackend` structure that regroup all the various backends.
 ```
 
-There are currently two ways to modify instantiated backends. The first one is to instantiate a new `ADModelBackend` and use `set_adbackend!` to modify `nlp`.
+To instantiate a new `ADModelBackend` while preserving existing backends, use `set_adbackend`.
 
 ```@example adnlp2
 adback = ADNLPModels.ADModelBackend(nlp.meta.nvar, nlp.f, gradient_backend = ADNLPModels.ForwardDiffADGradient)
-set_adbackend!(nlp, adback)
-get_adbackend(nlp)
+new_nlp = set_adbackend(nlp, adback)
+get_adbackend(new_nlp)
 ```
 
-The alternative is to use `set_adbackend!` and pass the new backends via `kwargs`. In the second approach, it is possible to pass either the type of the desired backend or an instance as shown below.
+An alternative way to use `set_adbackend` is to pass the new backends as keyword arguments.
+In this approach, you can pass either the type of the desired backend or an existing instance, as shown below.
 
 ```@example adnlp2
-set_adbackend!(
+new_nlp = set_adbackend(
   nlp,
   gradient_backend = ADNLPModels.ForwardDiffADGradient,
   jtprod_backend = ADNLPModels.GenericForwardDiffADJtprod(),
 )
-get_adbackend(nlp)
+get_adbackend(new_nlp)
 ```
 
-### Support multiple precision without having to recreate the model
+### Multi-precision model creation with backend reuse
 
 One of the strength of `ADNLPModels.jl` is the type flexibility. Let's assume, we first instantiate an `ADNLPModel` with a `Float64` initial guess.
 
@@ -133,11 +134,11 @@ x64 = rand(2)
 grad(nlp, x64)
 ```
 
-It is now possible to move to a different type, for instance `Float32`, while keeping the instance `nlp`.
+It is now possible to move to a different type for the gradient, for instance `Float32`, while keeping the other backends from the original model `nlp`.
 
 ```@example adnlp3
 x0_32 = ones(Float32, 2)
-set_adbackend!(nlp, gradient_backend = ADNLPModels.ForwardDiffADGradient, x0 = x0_32)
+new_nlp = set_adbackend(nlp, gradient_backend = ADNLPModels.ForwardDiffADGradient, x0 = x0_32)
 x32 = rand(Float32, 2)
-grad(nlp, x32)
+grad(new_nlp, x32)
 ```


src/ADNLPModels.jl
```diff
@@ -244,22 +244,25 @@ Returns the value `adbackend` from nlp.
...
 end
-function set_adbackend!(nlp::ADModel; kwargs...)
+
+function set_adbackend(nlp::ADModel; kwargs...)
   args = []
   for field in fieldnames(ADNLPModels.ADModelBackend)
     push!(args, if field in keys(kwargs) && typeof(kwargs[field]) <: ADBackend
       kwargs[field]
-    elseif field in keys(kwargs) && typeof(kwargs[field]) <: DataType
+    elseif field in keys(kwargs) && typeof(kwargs[field]) <: Union{DataType, UnionAll}
```
amontoison, line 265: @MaxenceGollier Why do you need `UnionAll`?

src/nlp.jl
```diff
@@ -1,18 +1,18 @@
 export ADNLPModel, ADNLPModel!
 
-mutable struct ADNLPModel{T, S, Si} <: AbstractADNLPModel{T, S}
+mutable struct ADNLPModel{T, S, Si, F1, F2 <: Function, ADMB <: ADModelBackend} <: AbstractADNLPModel{T, S}
```
amontoison, line 3:
```suggestion
mutable struct ADNLPModel{T, S, Si, F1, F2, ADMB <: ADModelBackend} <: AbstractADNLPModel{T, S}
```

src/nlp.jl
```diff
@@ -127,7 +127,7 @@ function ADNLPModel(f, x0::S; name::String = "Generic", minimize::Bool = true, k
   meta =
     NLPModelMeta{T, S}(nvar, x0 = x0, nnzh = nnzh, minimize = minimize, islp = false, name = name)
 
-  return ADNLPModel(meta, Counters(), adbackend, f, x -> T[])
+  return ADNLPModel(meta, Counters(), adbackend, f, (c, x) -> S[])
```
amontoison, line 130:
```suggestion
  return ADNLPModel(meta, Counters(), adbackend, f, (c, x) -> similar(S, 0))
```

src/nlp.jl
```diff
@@ -222,8 +222,7 @@ function ADNLPModel(
   ucon::S;
   kwargs...,
 ) where {S}
-  T = eltype(S)
-  return ADNLPModel(f, x0, clinrows, clincols, clinvals, x -> T[], lcon, ucon; kwargs...)
+  return ADNLPModel(f, x0, clinrows, clincols, clinvals, x -> S[], lcon, ucon; kwargs...)
```
amontoison, line 225:
```suggestion
  return ADNLPModel(f, x0, clinrows, clincols, clinvals, x -> similar(S, 0), lcon, ucon; kwargs...)
```

src/nlp.jl
```diff
@@ -348,7 +346,7 @@ function ADNLPModel(
     clinrows,
     clincols,
     clinvals,
-    x -> T[],
+    x -> S[],
```
amontoison, line 349:
```suggestion
    x -> similar(S, 0),
```

src/ADNLPModels.jl
MaxenceGollier: Because the type `ReverseDiffADGradient` or `ForwardDiffADGradient` are now parametric types and it is no longer true that `ReverseDiffADGradient <: DataType`...

amontoison: @tmigot Do you remember why you tested `typeof(kwargs[field]) <: DataType` ?
The condition `typeof(kwargs[field]) <: Type` is always `true` so I don't know why we are testing that.

amontoison:
@tmigot Are you fine with the new `set_adbackend` ?

src/ADNLPModels.jl
```diff
@@ -244,22 +244,25 @@ Returns the value `adbackend` from nlp.
...
 
-Replace the current `adbackend` value of nlp by `new_adbackend` or instantiate a new one with `kwargs`, see `ADModelBackend`.
+Create a copy of nlp that replaces the current `adbackend` with `new_adbackend` or instantiate a new one with `kwargs`, see `ADModelBackend`.
 By default, the setter with kwargs will reuse existing backends.
 """
-function set_adbackend!(nlp::ADModel, new_adbackend::ADModelBackend)
-  nlp.adbackend = new_adbackend
-  return nlp
+function set_adbackend(nlp::ADModel, new_adbackend::ADModelBackend)
+  T = typeof(nlp)
```
tmigot, line 254:
```suggestion
function set_adbackend(nlp::ADModel{T}, new_adbackend::ADModelBackend) where {T}
```

src/ADNLPModels.jl
tmigot: Yup, it doesn't make sense as is. Maybe it should be `kwargs[field] <: Type` to make sure the constructor call next make sense.

src/ADNLPModels.jl
```diff
@@ -244,22 +244,25 @@ Returns the value `adbackend` from nlp.
...
+  values = [f == :adbackend ? new_adbackend : getfield(nlp, f) for f in fieldnames(T)]
+  base_type = Base.typename(T).wrapper
+  return base_type(values...)
 end
-function set_adbackend!(nlp::ADModel; kwargs...)
+
+function set_adbackend(nlp::ADModel; kwargs...)
```
tmigot, line 260: I think this function should be called `ADNLPModel` and not `set_adbackend`, because it sounds more like a constructor than a setter. It can be another name too, but not `set...`

src/nlp.jl
```diff
@@ -127,7 +127,7 @@ function ADNLPModel(f, x0::S; name::String = "Generic", minimize::Bool = true, k
   meta =
     NLPModelMeta{T, S}(nvar, x0 = x0, nnzh = nnzh, minimize = minimize, islp = false, name = name)
 
-  return ADNLPModel(meta, Counters(), adbackend, f, x -> T[])
+  return ADNLPModel(meta, Counters(), adbackend, f, (c, x) -> similar(S, 0))
```
tmigot, line 130: Yes!

tmigot (changes requested):

test/utils.jl
```diff
@@ -12,25 +12,34 @@ ReverseDiffAD(nvar, f) = ADNLPModels.ADModelBackend(
...
-  @test typeof(get_adbackend(nlp).hprod_backend) <: ADNLPModels.ReverseDiffADHvprod
-  @test typeof(get_adbackend(nlp).jtprod_backend) <: ADNLPModels.GenericForwardDiffADJtprod
-  @test typeof(get_adbackend(nlp).hessian_backend) <: ADNLPModels.ReverseDiffADHessian
+  @test typeof(get_adbackend(newer_nlp).gradient_backend) <: ADNLPModels.ForwardDiffADGradient
+  @test typeof(get_adbackend(newer_nlp).hprod_backend) <: ADNLPModels.ReverseDiffADHvprod
+  @test typeof(get_adbackend(newer_nlp).jtprod_backend) <: ADNLPModels.GenericForwardDiffADJtprod
+  @test typeof(get_adbackend(newer_nlp).hessian_backend) <: ADNLPModels.ReverseDiffADHessian
+end
+
+function test_allocations(nlp)
```
tmigot, line 38: This function should be specialized to NLS problems to test `residual!` as well

tmigot:
@MaxenceGollier @amontoison I moved the PR from draft to review to run the package benchmark by curiosity

MaxenceGollier:
> Why the tests are failing? Is it just because NLS grad! function uses other function behind and should not be tested or is there some typing issue in NLPModels?

That's what I have been trying to figure out, some Hessian (with JET I mean) tests also seem to fail for some reason so there might be other unstabilities hidden in this repo...
> what were the tests problems?

What do you mean ? as I said, I still have to figure things out, I have a lot of urgent work to do on something else next week, I will be available to work this out the week after that...

> Can you create an issue in NLPModelsTest.jl to add this in all models?

Yes good idea, I found similar issues on `QuadraticModels.jl`... (still need to open an issue but I don't have much time currently)

src/ADNLPModels.jl
MaxenceGollier: `typeof(nlp) != T where nlp::ADModel{T}`. I added something else instead which i think is what you meant :)

MaxenceGollier:
What do you think of this version of `set_adbackend` ? @amontoison @tmigot. I tried to implement all your suggestions

MaxenceGollier:
I think tests keep failing due to inherent type instabilities in `ForwardDiff.jl`. See in [ForwardDiff/docs/advanced.md](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/docs/src/user/advanced.md) : 

> If your input dimension is constant across calls, you should explicitly select a chunk size rather than relying on ForwardDiff's heuristic. There are two reasons for this. The first is that ForwardDiff's heuristic depends only on the input dimension, whereas in reality the optimal chunk size will also depend on the target function. The second is that ForwardDiff's heuristic is inherently type-unstable, which can cause the entire call to be type-unstable.

The remaining type instabilities should be solved by having a closer look to those of `ForwardDiff`.

amontoison:
@tmigot 
Can you take care of the review? I am still traveling this week and won't be able to have a look before 2 weeks at least.

test/Project.toml
```diff
@@ -1,5 +1,6 @@
 [deps]
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
```
tmigot, line 3: Add a version for JET (I think even the packages from Base have versions now)

test/utils.jl
```diff
@@ -12,25 +12,53 @@ ReverseDiffAD(nvar, f) = ADNLPModels.ADModelBackend(
...
+end
+
+function test_allocations(nlp::ADNLSModel)
+  x = nlp.meta.x0
+  y = zeros(eltype(nlp.meta.x0), nlp.meta.ncon) 
+  g = zeros(eltype(nlp.meta.x0), nlp.meta.nvar)
+  Fx = zeros(eltype(nlp.meta.x0), nlp.nls_meta.nequ)
+  @test_opt target_modules=(ADNLPModels,) obj(nlp, x)  
+  @test_opt target_modules=(ADNLPModels,) cons!(nlp, x, y)
+  @test_opt target_modules=(ADNLPModels,) grad!(nlp, x, g, Fx)
```
tmigot, line 63:
```suggestion
  @test_opt target_modules=(ADNLPModels,) grad!(nlp, x, g, Fx)
  @test_opt target_modules=(ADNLPModels,) residual!(nlp, x, Fx)
```

src/forward.jl
```diff
@@ -108,8 +108,8 @@ function GenericForwardDiffADJtprod(
 )
   return GenericForwardDiffADJtprod()
 end
-function Jtprod!(::GenericForwardDiffADJtprod, Jtv, f, x, v, ::Val)
-  Jtv .= ForwardDiff.gradient(x -> dot(f(x), v), x)
+function Jtprod!(::GenericForwardDiffADJtprod, Jtv::S, f, x::S, v::S, ::Val) where{S}
+  Jtv .= ForwardDiff.gradient(x -> dot(f(x), v), x)::S
```
tmigot, line 112: This would be hiding the issues I think. `Jtv` is of type `S`, so we shouldn't have to convert it to `S` again. If that's so, it indicates that there are some issues and it probably should not be done silently.
We then need to either add a warning or/and add a unit test + open an issue to track this.
```suggestion
  Jtv .= ForwardDiff.gradient(x -> dot(f(x), v), x)
```

src/forward.jl
```diff
@@ -151,7 +151,7 @@ function Jtprod!(b::ForwardDiffADJtprod{Tag, GT, S}, Jtv, c!, x, v, ::Val) where
   b.sol[1:ncon] .= 0
   b.sol[(ncon + 1):(ncon + nvar)] .= x
   b.sol[(ncon + nvar + 1):(2 * ncon + nvar)] .= v
-  ForwardDiff.gradient!(b.temp, b.ψ, b.sol, b.cfg)
+  ForwardDiff.gradient!(b.temp, b.ψ, b.sol)
```
tmigot, line 154: This doesn't seem right.

src/forward.jl
```diff
@@ -138,8 +138,8 @@ function ForwardDiffADJtprod(
     c!(cx, x)
     dot(cx, u)
   end
-  tagψ = ForwardDiff.Tag(ψ, T)
-  cfg = ForwardDiff.GradientConfig(ψ, temp, ForwardDiff.Chunk(temp), tagψ)
+  #tagψ = ForwardDiff.Tag(ψ, T)
+  cfg = ForwardDiff.GradientConfig(ψ, temp)
```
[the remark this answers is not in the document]
tmigot: We should probably investigate this further, because there was a performance reason we did this.

src/ADNLPModels.jl
```diff
@@ -244,22 +244,24 @@ Returns the value `adbackend` from nlp.
...
+    new_nlp = set_adbackend(nlp; kwargs...)
 
-Replace the current `adbackend` value of nlp by `new_adbackend` or instantiate a new one with `kwargs`, see `ADModelBackend`.
+Create a copy of nlp that replaces the current `adbackend` with `new_adbackend` or instantiate a new one with `kwargs`, see `ADModelBackend`.
 By default, the setter with kwargs will reuse existing backends.
 """
-function set_adbackend!(nlp::ADModel, new_adbackend::ADModelBackend)
-  nlp.adbackend = new_adbackend
-  return nlp
+function _set_adbackend(nlp::ADM, new_adbackend::ADModelBackend) where{ADM}
```
tmigot, line 253:
```suggestion
function _set_adbackend(nlp::ADM, new_adbackend::ADModelBackend) where{ADM <: AbstractADNLPModel}
```
otherwise nlp.f might lead to an error.

src/ADNLPModels.jl
```diff
@@ -269,8 +271,23 @@ function set_adbackend!(nlp::ADModel; kwargs...)
...
+    return _set_adbackend(nlp; kwargs...)
+end
+
+function ADNLSModel(nlp::ADNLSModel; kwargs...)
+    return _set_adbackend(nlp; kwargs...)
+end
+
+function ADNLSModel(nlp::ADNLSModel, new_adbackend::ADModelBackend)
+    return _set_adbackend(nlp, new_adbackend)
 end
```
tmigot, line 291: Good idea, but it doesn't work with this https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/fd0d2489efdee9ef4c0f0e139e7c3864b3c543b9/src/ADNLPModels.jl#L35 .
Because, we are already building mixed models.

Do we really need the ADNLPModel(nlp::ADNLPModel; kwargs...) and ADNLSModel(nlp::ADNLSModel; kwargs...) ?

src/forward.jl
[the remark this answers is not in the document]
tmigot: Do you have a MWE or did you observe this via the tests?

tmigot (changes requested):
Thanks a lot @MaxenceGollier for the great work! I made some comments, but we are going in a good direction.

test/utils.jl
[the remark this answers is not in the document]
tmigot: @MaxenceGollier can you put the link here once it's done, thanks!

tmigot:
> I think tests keep failing due to inherent type instabilities in `ForwardDiff.jl`. See in [ForwardDiff/docs/advanced.md](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/docs/src/user/advanced.md) :
> 
> > If your input dimension is constant across calls, you should explicitly select a chunk size rather than relying on ForwardDiff's heuristic. There are two reasons for this. The first is that ForwardDiff's heuristic depends only on the input dimension, whereas in reality the optimal chunk size will also depend on the target function. The second is that ForwardDiff's heuristic is inherently type-unstable, which can cause the entire call to be type-unstable.
> 
> The remaining type instabilities should be solved by having a closer look to those of `ForwardDiff`.

I am unfortunately not too surprised by this. Thanks to your analysis we will improve our part of the code.
We should try to think of a way to still test that our code is good.

amontoison:
bump

MaxenceGollier:
> I am unfortunately not too surprised by this. Thanks to your analysis we will improve our part of the code. We should try to think of a way to still test that our code is good.

I agree. Perhaps the best we can do is to call the `report_opt` macro but let it ignore all the issues from the `ForwardDiff` package.
If our code is stable with all other backends then we'd consider that our code is good enough.

src/forward.jl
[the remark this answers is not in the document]
MaxenceGollier: Just via the tests. I am reverting these changes.

amontoison:
We need to open another PR to fix the benchmarks with Julia 1.12.
Once it is merged, we can rebase this branch on top of it.
I can't have a look before Sunday / next week if you don't do it before.

test/Project.toml
```diff
@@ -12,6 +13,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 ForwardDiff = "0.10"
+JET = "0.9, 0.10"
```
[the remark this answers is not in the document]
amontoison: No, you can run any test on LTS otherwise.
JET `0.10` requires Julia 1.12 as a compat entry.

amontoison:
@tmigot We should be consistent for Maxence and don't give contrary comments. I think a non-inplace `set_backend` is still relevant and return a new ADNLPModel.

test/Project.toml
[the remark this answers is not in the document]
tmigot: Oh, I see. Okay then. Thanks!

tmigot:
Sounds, good.

MaxenceGollier:
Hi guys, I think we should move forward with this, I believe this is an important improvement for this package.
As discussed above, I need to have the commits in a branch of this repo and not a local fork for the benchmarks to run but i don't have rights to even create a branch.

Perhaps we could
1. Create a branch, say mgollier-type-stab-dev on this repo,
2. I make a PR from the branch in my fork with the commits to this new branch
3. We run the benchmarks from there
4. Open a final PR from the mgollier-type-stab-dev branch to the master branch with the updates in the docs etc.

Please let me know.

MaxenceGollier:
Superseded by #365.

PR #357, opened 2025-07-21
Disable gradient and Hessian backends for NLSModels

dpo:
A user can still enable them manually if they really want to.

### DIFF
--- a/test/manual.jl
+++ b/test/manual.jl
@@ -4,28 +4,13 @@ function test_nlp_consistency(nlp, model; counters = true)
   v = 2 * ones(nvar)
   y = ones(ncon)
 
-  @test grad(nlp, x) == grad(model, x)
-  @test !counters || (neval_grad(model) == 2)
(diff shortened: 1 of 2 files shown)


src/ad.jl
```diff
@@ -218,14 +218,14 @@ function ADModelNLSBackend(
...
   matrix_free::Bool = false,
   show_time::Bool = false,
-  gradient_backend = get_default_backend(:gradient_backend, backend),
-  hprod_backend = get_default_backend(:hprod_backend, backend),
-  hessian_backend = get_default_backend(:hessian_backend, backend, matrix_free),
-  hprod_residual_backend = get_default_backend(:hprod_residual_backend, backend),
+  gradient_backend = EmptyADbackend(),
+  hprod_backend = EmptyADbackend(),
+  hessian_backend = EmptyADbackend(),
+  hprod_residual_backend = EmptyADbackend(),
```
tmigot, line 224: Don't you need this one?

amontoison (approved):

